PR #1, opened 2016-11-11
Better reporter

okonet:
This is how it will look after the merge:

2016-11-14 at 12 36

but it's much easier to set up now by tweaking constants.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "logux-core": "logux/logux-core",
     "logux-sync": "logux/logux-sync",
     "short-uuid": "^2.1.0",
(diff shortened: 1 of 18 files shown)


okonet:
@ai I think this is solid and can be merged now. Thoughts?

ai:
Sure, I like latest style.

Could you fix Travis (maybe you need to update server snapshots too, same `./test/servers/update.js` way)?

okonet:
I was wondering why the build is failing. Just did that and tests are now green locally.

ai:
Awesome. I will merge it tonight (my open source time is after the main work). Thanks for your work!

PR #14, opened 2017-01-14
Update reporters test to use Jest snapshots

PavloKovalov:
Summary:
- Switched reporters test to Jest snapshots
- Added snapshot update command `yarn update-snapshots`


package.json
```diff
@@ -42,7 +42,8 @@
     "clean": "rimraf docs/ coverage/",
     "lint": "eslint *.js test/*.js",
     "docs": "jsdoc --configure .jsdocrc *.js",
-    "test": "jest --coverage && yarn run lint && yarn run spellcheck"
+    "test": "jest --coverage && yarn run lint && yarn run spellcheck",
+    "update-snapshots": "jest -u"
```
ai: I can run `jest -u` in console by my `n` function https://github.com/ai/environment/blob/master/zshrc#L52-L61

Could you remove it?

test/reporter.test.js
```diff
@@ -1,14 +1,57 @@
...
-var fs = require('fs')
-
-var promisify = require('../promisify')
 var reporter = require('../reporter')
-var reports = require('./snapshots/reports')
+
+function normalizeNewlines (string) {
+  // use local copy of Jest newline normalization function
+  // until Jest doens't apply normalization on comprasion
+  return string.replace(/\r\n|\r/g, '\n')
```
ai: BTW, reporter uses `\v\r` on UNIX, do we need to cover it here?

ai:
I added few comments. But in this PR I dodn’t found a way to check reporter output in console.

In previous test system I could call `./test/snapshots/update` and it print all snapshots. Without printing a snapshots I couldn’t check reporter output.

test/reporter.test.js
PavloKovalov: It covers only newlines normalized by Jest during writing snapshots.

package.json
PavloKovalov: removed

test/reporter.test.js
ai: So what lines do we have in the end? :) `\n` or `\v\n`?

PavloKovalov:
@ai can you clarify, what the exact purpose of separate tool for checking reporter output?
If output changes it will be shown as a diff on jest report with colors. All color charactes are saved in snapshot.
Also, it can be separate PR.

ai:
@PavloKovalov right now I use current snapshots for 2 purposes:

1. Regression tests.
2. Debug tool to see current reporters output.

For example, I want to change how Logux report about error. I need to see somewhere this output, right?

Try to run `test/snapshots/update`. It will not only update the snapshots, but also print latest snapshots to terminal (so I can check output with my eyes).

So we need this behavior for new testings system too. How I will check how output is look like?

test/reporter.test.js
PavloKovalov: hmm… didn't find `\v\n` in produced jest output, but i'm on macos.
need to dig on it.

ai: OK, let’s make print tool and we will see what symbols are in snapshots anyway

PR #15, opened 2017-01-23
Better errors descriptions for listen();

morozRed:
Summary:
- Add error-helper.js
- Add error-helper.test.js

### DIFF
--- a/server.js
+++ b/server.js
@@ -56,10 +56,10 @@ function Server (options) {
   var app = this
 
   function onError (e) {
-    if (e.syscall === 'listen') {
+    try {
       process.stderr.write(errorHelper(e, app))
-    } else {
-      app.reporter('runtimeError', app, undefined, e)
+    } catch (err) {
+      app.reporter('runtimeError', app, undefined, err)
     }
     app.destroy().then(function () {
       process.exit(1)
--- a/server.js
+++ b/server.js
@@ -55,7 +55,8 @@ function Server (options) {
   var app = this
 
   function onError (e) {
-    app.reporter('runtimeError', app, undefined, e)
+    e.syscall === 'listen' ? process.stderr.write(errorHelper(e, app)) :
+        app.reporter("runtimeError", app, undefined, e)
     app.destroy().then(function () {
       process.exit(1)
     })
--- a/server.js
+++ b/server.js
@@ -1,5 +1,6 @@
 var BaseServer = require('./base-server')
 var reporter = require('./reporter')
+var errorHelper = require('./error-helper')
 
 /**
  * End-user API to create Logux server.
@@ -55,8 +56,11 @@ function Server (options) {
   var app = this
 
   function onError (e) {
-    e.syscall === 'listen' ? process.stderr.write(errorHelper(e, app)) :
-        app.reporter("runtimeError", app, undefined, e)
+    if (e.syscall === 'listen') {
+      process.stderr.write(errorHelper(e, app))
+    } else {
+      app.reporter('runtimeError', app, undefined, e)
+    }
     app.destroy().then(function () {
       process.exit(1)
     })
--- a/error-helper.js
+++ b/error-helper.js
@@ -1,0 +1,82 @@
+var yyyymmdd = require('yyyy-mm-dd')
+var stripAnsi = require('strip-ansi')
+var chalk = require('chalk')
+var path = require('path')
+var os = require('os')
+
+var pkg = require('./package.json')
+
+var PADDING = '        '
+var SEPARATOR = os.EOL + os.EOL
+var NEXT_LINE = os.EOL === '\n' ? '\r\v' : os.EOL
+
+function rightPag(str, length) {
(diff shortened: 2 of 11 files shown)


ai:
Good start. But if we want good docs for new users we should have better error message.

Let’s think together. Why user will get `Port already in use` error? 31337 and 1337 ports are very rare. So it is very rare case that some other non-Logux program use this port.

In my experience there are 2 most popular cases:

1. User have 2 Logux projects. He worked on first. And now manager asked him to fix some issue in second project. And developer forget to stop first Logux server before run second Logux server.
2. You deploy new Logux app to servers. But previous app was not killed correctly because of error in deploy script.

In all that cases “Try to use other port” will make only worse.

So, let’s change error text. You could try start with `port in use case`. Describe (with simple language) what is the problem. Describe possible reason why it happens. For each reason describe how to solve it.

You can start here. Just write it in comment.

morozRed:
Thanks!
Ok, we can do something like this:
**EADDRINUSE**
```
ERROR   Port :3000 already in use.
HINT    Looks like Logux server already running on this port.
        Make sure there is no Logux servers running on this port.
```
**EACCES**
```
ERROR  You are not allowed to run server on this port.
HINT   Try to run server on port > 1024.
       You are currently using port: ${port}.
```

ai:
Yeap, much better. But let’s write more details:

```
ERROR   Port :3000 already in use.
        Another Logux server or other app already running on this port.
        Maybe you didn’t not stop server from other project
        or previous version of this server was not killed.
```

Now you need to do same work with `EACCES`. Explain user why `<= 1024` posrt are different. What is the main reason of `EACCES` error?

morozRed:
What about this?
**EACCES**
```
ERROR  You are not allowed to run server on this port.
HINT   You are not allowed to run your server on ports <= 1024
       Because they are privileged by security reasons.
       Try to run your server on ports > 1024.
```

ai:
@morozRed do write docs without analysis ;). Tell me possible scenarios when users will have `EACCES` error.

morozRed:
If we are talking only about listen() exceptions
1. User did something like like this in his app 
```
 app.listen({
    port: 1000
}) 
```
2. User is trying to access a file in a way forbidden by its file access permissions

ai:
1. Why he or she will do it? I think most of users will copy code from docs ;).
2. Other reason — developer forget to write `sudo` to start server. I had this mistake for my own too :D.
3. `EACCES` could be for files too?!!! In this case we should detect port and file `EACCES`.

morozRed:
Yes, you I just found it here: https://nodejs.org/api/errors.html#errors_error_code
I'll found a way we can detect EACCES type :) And then we will discuss error ` HINT ` again

morozRed:
After little investigation I've realised that ` EACCES ` for files will not appear on listen() call. So we can finalize error message.
```
ERROR  You are not allowed to run server on this port
       Try to run server with sudo prefix or use port >= 1024
```

ai:
Listen could load certificate files for TLS. Maybe we need good message for them.

morozRed:
I've tested it and got this error:
``` 
ERROR  EACCES: permission denied, open '/server.key' at 2017-01-24 17:23:42

{ Error: EACCES: permission denied, open '/server.key'
  errno: -13,
  code: 'EACCES',
  syscall: 'open',
  path: '/server.key' }
```

ai:
We need to cover this case too with good instructions.

ai:
By the way, if `try to add sudo` is good recommendation. Running server as root is very very dangerous. Maybe we should ask “try to change user”?

morozRed:
Ok, I'll think about this case with certificate.
Yes, probably it's better.

ai:
OK, now let’s update text in PR and I will accept it.

ai:
Few things to fix in code:

1. Use some real port instead of `undefined` in tests.
2. Move common code from `error-helper` and `reporter` to some common file.
3. You don’t need tricky `SEPARATOR` and `NEW_LINE` hack in `error-helper`.
4. Add real integration test. You can use some examples from `test/servers.js` tests.
5. Instead of `e.syscall === 'listen'` hack just use `try-catch`. Simple solutions is always better.

morozRed:
Updates:
- updated errors messages
- add tests
- fix old tests
- add log-helper.js

server.js
```diff
@@ -55,7 +56,11 @@ function Server (options) {
   var app = this
 
   function onError (e) {
-    app.reporter('runtimeError', app, undefined, e)
+    try {
+      process.stderr.write(errorHelper(e, app))
+    } catch (err) {
+      app.reporter('runtimeError', app, undefined, err)
+    }
```
ai: You misunderstand me. `onError` should contains only `app.reporter('runtimeError', app, undefined, e)`.

I told you to write try-catch in `listen()` method:

```
listen: function listen () {
  try {
    return BaseServer.prototype.listen.apply(this, arguments)
  } catch (e) {
    process.stderr.write(errorHelper(e, app))
  }
}
```

PR #19, opened 2017-02-12
ES6 syntax and eslint config update

lamartire:


ai:
Could you rebase your PR to `logux/eslint-config-logux` master, because @frsv already updated it to ES6.

Also we need a addition rule to force `arg => …`, instead of `(arg) => …`. @frsv maybe you know it  and could send PR to config?

frsv:
Actually, we already have this rule here https://github.com/logux/eslint-config-logux/blob/master/node4.js#L20

ai:
@frsv great :D

ai:
@lamartire don’t forget to fix all Travis CI issues ;)

lamartire:
@ai can you update npm-package of `logux/eslint-config-logux`?

ai:
I plan to release it after we will be sure in it (move Logux Server to it).

Why you need a npm release?

lamartire:
@ai after installig  dependencies I'll take old version and eslint not worked.
Or can I install it manually from repository?

lamartire:
Add some rules for templates -- https://github.com/logux/eslint-config-logux/pull/5

lamartire:
Need some help with tests.
Async test broken:



Tried solutions from docks, but it did not help.

ai:
Yarn has a problem with updating package from git. Just run `rm -R node_modules && yarn cache clean && yarn` to be sure. In Travis CI I just clean cache and run task again.

ai:
Async task is broken because you change it. Don’t use `done` in Jest, just return a Promise.

lamartire:
New problems.
`SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode`, need update config again, besause we use `'strict': ['error', 'global']`.

ai:
`SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode` is not from ESLint, it is directly from node.js ;)

lamartire:
So, I must remove all rules with `let`, `const` and keep all "var-rules"?

ai:
Yeap, replace `let` and `const` back to `var`.

ai:
Don't change a docs. Developers mostly will run server on latest node. And developers stocked on node 4 know about syntax limitation. This is why I prefer to force node 4 users to convert syntax from docs example, rather than force node 6 users to convert syntax to latest ES2016.

lamartire:
I think the remaining errors relate to arrow functions. @ai, does it possible reason?

ai:
Also there is very a `euse` typo. It should be `use`.

ai:
Yeap. All Jest test are executed in parallel. So I used `this.app` to set current server for each year separated. Good question how to deal with it without `this`. If you have no ideas too, ask Jest author.

lamartire:
I returned documentation and other files. Misstake with glob :cry:
I think I will open issue in jest tomorrow, because I don't have ideas now.

ai:
Sure. We did a great work. I think `this` issue will be the last problem.

frsv:
@lamartire I did some workaround to achieve normal operation with this context while working with another issue. You can try to do something similar – https://github.com/frsv/logux-server/commit/00016d1cd7f703df786d24a858371fb42518213a#diff-7786459c30ea1024c9ec88dbd9a1f380R47

But, maybe I'll renounce this solution finally

lamartire:
@ai what about option, like "runOnFreePort" or something like this.
When default port are used, logux try to find free port and notifies user when it was found.

ai:
We can just add function `nextPort` in test, which will just return next port (2000, 2001, 2002, …)

lamartire:
So, then we must check port before test running, right?

ai:
Honestly, we could do it without a checking that ports are really free.

ai:
Just take some rare numbers.

lamartire:
I fixed many issues with ports, but have fails.
There are still some problems, which are reproduced locally on master.

ai:
How do you stop server in `afterEach` without a `this`?

ai:
With `function () { }` `this` in tests means current test context. With `() =>` it means global context.

So you have all this problems, because you didn't stop server after a test.

This is why we should write to Jest authors and ask them about API to access to test context with ES6 arrow functions.

lamartire:
Clear.
Do you mind if I quote the comment about arrows in jest issue?
But on the one hand, I can try solve problem with save context in variable. Like @frsv advised.

ai:
Could you show a that variable based solution here in small example (I don't have laptop right now)

lamartire:
I'm not sure, but:

```js
var ctx;

beforeEach(function() {
  var ctx = this;
})

//...

afterEach(function() {
  if (ctx && ctx.server) {
     ctx.server.close()
     ctx = null;
  }
})
```

lamartire:
@frsv example -- https://github.com/frsv/logux-server/commit/00016d1cd7f703df786d24a858371fb42518213a#diff-7786459c30ea1024c9ec88dbd9a1f380R47

ai:
Every test executed in parallel in Jest. This context variable in global, so it will not work 😉 It is just finally variable, `beforeEach` didn't do anything.

frsv:
@ai but why it works locally? In the end of each test server shuts down.

ai:
@frsv concurrency is tricky thing. Anyway current code is wrong, we need to fix it. Please, ask Jest authors and post a link to issue here.

frsv:
Well, @lamartire has created issue already – https://github.com/facebook/jest/issues/2883

ai:
@frsv great! 

@lamartire also don’t forget to mention somebody from core team. It increase response time.

lamartire:
@ai thank for advice 😸

ai:
OK, seems like task doesn’t go in parallel anymore. Let’s replace `this.app` to just `app` as in comment https://github.com/facebook/jest/issues/2883#issuecomment-279914292

And them I will see a issues.

lamartire:
Another way - https://github.com/facebook/jest/issues/2883#issuecomment-279960943
How about ignore this test and use it unmodified if replacing `this.app` by `app` will fail?

ai:
Replace `this.app` to `app` and then I will look and fix them.

lamartire:
I think I can try resolve remaining issues myself today.
If not, I write here.

lamartire:
Ha, all tests done! :smile_cat:

ai:
Wow! Great.

ai:
@frsv @lamartire I plan to split cult of martians task to two to mark both of you as winners. Is it OK?

ai:
@lamartire I need your VK and Twitter accounts

ai:
@lamartire you still miss `this` and `() =>` :D read about arrow-functions more careful.

This code will not work as you expected:

```
beforeEach(() => {
  var testId = shortid.generate()
  this.currentTest = testId
  this.tests[testId] = {}
})
```

`this` will be equal to `global`. So there is not point to have `this.tests`. You could do it simpler:

```js
var currentTest
var tests = { }

beforeEach(() => {
  currentTest = shortid.generate()
  tests[testId] = {}
})
```

lamartire:
All right. Thanks @ai! :)
I have not twitter, can give you my VK by email.

ai:
@lamartire this VK is needed to announce. If you want to keep it private I can just mentioned you by name :)

lamartire:
I have [empty twitter](https://twitter.com/eritramal) and little more filled vk. Let me be just Kostya, before another time when my skills will be better 😸

PR #20, opened 2017-02-19
Client could not get app's nodeId

Yevs:
Client could not get app's nodeId because of a typo `app.nodeId` instead of `app.options.nodeId`


ai:
Wow, great. First bugfix PR :D.

PR #23, opened 2017-03-16
Sql store

gazay:
Store tested with postgresql, other DBs coming soon, but should be same as we don't use here special features for postgre.

Coverage is 99.2% as I can't figure out how to test this line https://github.com/logux/logux-server/compare/master...gazay:sql-store?expand=1#diff-36cbc6fdaa55557b7b013f0cdbabcce9R70

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,8 @@ node_js:
   - stable
   - "6"
   - "4"
+
+addons:
+  postgresql: '9.4'
+before_script:
+  - psql -c 'create database logux_test;' -U postgres
--- a/.yaspellerrc
+++ b/.yaspellerrc
@@ -33,6 +33,12 @@
     "stacktrace",
     "debugError",
     "nanoevents",
-    "SQLStore"
+    "SQLStore",
+    "precreated",
+    "postgres",
+    "sqlite",
+    "mssql",
+    "logux",
+    "sequelize"
   ]
 }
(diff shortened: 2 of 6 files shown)


gazay:
Coverage is ok now.

ai:
Great work. Few notes:

1. Does any Logux Server contributor will need a PostgreSQL to run a tests right now? Maybe we could add SQLite driver by default and run PostgreSQL only on Travis CI (replace driver by environment variable)?
2. I see `currentOffset` in `nextEntry`. Will it work correctly if somebody will add new log entry between `nextEntry` calls (because it will change offset)?
3. We should avoid nested `then()`. Promises was created to make code clean by replacing nested callbacks to linear. But, of course, 100% linear `then()` is not always possible, so I am not sure about this case.
4. Should we use `minAdded`/`maxAdded` in `WHERE` in `removeReason()`.
5. If `username` and `password` could be empty, maybe we could move them to `options` object? It will useful when you want to change `dialect`, but don’t use username/password — so you will not need to write `null, null`. Also `sqlite` doesn’t have username/password at all. In other had, with this API we copy `Sequelize` API, but I am not sure, that it is the best API :D.
6. In Logux Server we moved to ES6 syntax, so you could use ES6 classes syntax (much better that in ES5). Check example in `base-server.js`.

ai:
BTW, anyway I will not have release at React London (servers channels is main problem), so we could finish it without rush.

Issue #26, opened 2017-03-20
Unstable acceptance test 'shows uncatch rejects'


artmnv:
Acceptance test 'shows uncatch rejects' in server.test.js has this snapshot:
```
 ERROR  Test Error at 1970-01-01 00:00:00
        fake stacktrace

Error event: Test Error
 INFO   Shutting down Logux server at 1970-01-01 00:00:00
```

But sometimes two last lines come in another order. That's because of race condition. We need to stabilise this test. Maybe with some timeout?

artmnv:
Example of failed job: https://travis-ci.org/logux/logux-server/jobs/213092586

ai:
Yeap, seems like tests are not 100% stable. Do you want to send PR?

artmnv:
@ai If I'll figure out how to fix this. Any hits? Anyway I'll think for myself.

ai:
Have no idea right now. I can help you here after React London (March 29).

PR #28, opened 2017-03-25
Removing of uncatch test uncertainty

artmnv:
Issue #26 

I made research and this solution works. There are two main things:
1. We are making timeout for destruction start message come first.
2. We are stoping server from total destruction until error message shows up.

### DIFF
--- a/test/__snapshots__/server.test.js.snap
+++ b/test/__snapshots__/server.test.js.snap
@@ -81,9 +81,9 @@ exports[`shows uncatch rejects 1`] = `
 " ERROR  Test Error at 1970-01-01 00:00:00
         fake stacktrace
 
-Error event: Test Error
  INFO   Shutting down Logux server at 1970-01-01 00:00:00
 
+Error event: Test Error
 "
 `;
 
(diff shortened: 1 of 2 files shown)


artmnv:
I don't like this solution. Timeouts in tests are bad thing. But it works. This solution is better than unstable test. But it's better to think about more stable tests paradigm and logux architecture which can be tested without timeouts.

ai:
Why we need `console.log` in diff? Server should output error message by itself (we need to check that server prints errors).

artmnv:
@ai There was `console.log` in this test before. I thought the meaning of it is that we want to check if we can add any custom callback on error with `app.on('error', callback)`. So we'll get not only default server error message but custom action too. Am I wrong?

ai:
I worry that we change test too much. Users will not have this `errorPromise`.

Maybe we should increase `new Promise((resolve, reject) => {` (origin one) timeout 50 → 100?

I agree that timeout based testing is not good :(. But it is hard to test CLI without them. JS API could be tested without timeouts more easy, than CLI.

Issue #41, opened 2017-08-07
Allow to provide a reporter to the Server


ertrzyiks:
Hi,

I'm trying to add tests for my server class and to do that I need to pass a reporter mock. It's possible only with `BaseServer`, but `Server` class builds reporter on it's own.

```js
class MyServer extends Server {
 ...
}

new MyServer(options, reporterMock) // it doesn't work
```

The solution could be to inherit from `BaseServer` and copy-paste the bootstrap code from `Server`

```js
class MyServer extends BaseServer {
 ...copy paste Server code here
}

new MyServer(options, reporterMock) // it works!
```

What do you think about adding an option to override reporter in Server class as well? Or maybe compose Server with BaseServer instead of inherit from it?

ai:
I like the idea too override reporter in Server, good idea.

Please send PR.

PR #43, opened 2017-08-07
Allow to override reporter in Server

ertrzyiks:
Related to #41

I don't see any specs for this file yet, I tested it manually on my machine. Is it enough or you want some specs for this too?

### DIFF
--- a/server.js
+++ b/server.js
@@ -139,10 +139,10 @@ class Server extends BaseServer {
     return options
   }
 
-  constructor (options) {
+  constructor (options, reporter) {
     if (!options) options = { }
     options.reporter = options.reporter || 'human'
-    const reporter = createReporter(options)
+    if (!reporter) reporter = createReporter(options)
 
     let initialized = false
     const onError = err => {
--- a/server.js
+++ b/server.js
@@ -142,12 +142,9 @@ class Server extends BaseServer {
   constructor (options) {
     if (!options) options = { }
 
-    let reporter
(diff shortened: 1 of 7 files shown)


ai:
Unfortunately, we need tests for it.

Server tests looks tricky, but it is not hard. Create a new server with custom reporter at `test/servers/` and check their output by snapshots in `server.test.js`.

ertrzyiks:
This was also my impression that specs are needed here, I was asking just in case. I will add them tomorrow 👍

ai:
Also maybe we should pass `reporter` in `options` object? 

I used the second option in `BaseServer` only because `reporter` was some sort private option. If we make in public, we can add it to `options` in `BaseServer` too.

server.js
```diff
@@ -141,8 +141,11 @@ class Server extends BaseServer {
 
   constructor (options) {
     if (!options) options = { }
-    options.reporter = options.reporter || 'human'
-    const reporter = createReporter(options)
+
+    if (typeof options.reporter !== 'function') {
+      options.reporter = options.reporter || 'human'
+      options.reporter = createReporter(options)
```
ertrzyiks: I don't like it to be honest, maybe something like?

```
createReporter(type, options)
```

ertrzyiks:
@ai Actually, it was easier than expected. There were specs for this class already, I missed them at first glance.

What do you think?

ai:
Great PR. I especially like your idea to use `reporter` for functions too.

Issue #55, opened 2019-04-09
Use async/await


ai:
Logux Server 0.3 requires Node.js >= 10, so we are safe to use `async`/`await` instead of `Promise`

ai:
Part of the works was done in https://github.com/logux/server/pull/58

@chafilin can you show me links to the code, where you have problems in replacing Promise?

ai:
@chafilin don’t forget about tests. They are the main users of `.then`.

chafilin:
> Part of the works was done in #58
> 
> @chafilin can you show me links to the code, where you have problems in replacing Promise?

This kind of Promises: https://github.com/logux/server/blob/b47dccd9e0093a4a8dc65fa06195046c2e2d1462/bind-backend-proxy.js#L106
https://github.com/logux/server/blob/b47dccd9e0093a4a8dc65fa06195046c2e2d1462/bind-backend-proxy.js#L42

ai:
Yeap. Let’s keep Promise in those 2 examples.

I am waiting for PR to replace `then` in tests.

ai:
The current list of `.then` in project

```
➜ rg "\.then"
server-client.js
275:      .then(result => {

base-server.js
388:        .then(() => Promise.all(before))
389:        .then(keys => new Promise((resolve, reject) => {
405:    promise = promise.then(() => {
410:      promise.then(() => {
420:    return promise.then(() => {
424:    }).then(() => {
921:        }).then(access => {
960:            }).then(() => emitSubscribed())
964:        }).then(access => {

force-promise.js
8:  if (typeof result !== 'object' || typeof result.then !== 'function') {

test/force-promise.test.js
4:  return forcePromise(() => Promise.resolve('result')).then(result => {
11:  return forcePromise(() => Promise.resolve().then(() => {
19:  return forcePromise(() => 'result').then(result => {

test/base-server.test.js
580:  app.destroy().then(() => {

test/stress/client.js
70:  delay(Math.random() * 10000).then(() => {

test/stress/backend.js
46:    let processing = delay(500).then(() => {
50:      processing = processing.then(() => {
52:      }).then(() => {
58:      }).then(() => {
62:      processing = processing.then(() => {
64:      }).then(() => {
68:    processing.then(() => {
```

Seems like we have only a few lines to close this issue.

ai:
https://github.com/logux/server/commit/a369b484acd10bd1697c9345eff10bc78689886c
[...]

PR #58, opened 2019-04-24
Refactor to async/await

chafilin:
Issue #55 

1. Changed all `.then` constuctions to async/await. Except `base-server.js` and filter method in `server-client.js`
2. Did't touch force-promise, because after refactoring whole code it will be useless
3. Don't know how to refactor `promisify` and `retrurn new Promise` with wrapping callback

### DIFF
--- a/server.js
+++ b/server.js
@@ -252,8 +252,7 @@ class Server extends BaseServer {
 
   async listen (...args) {
     try {
-      let result = await BaseServer.prototype.listen.apply(this, args)
-      return result
+      return await BaseServer.prototype.listen.apply(this, args)
     } catch (err) {
       this.reporter('error', { err })
       process.exit(1)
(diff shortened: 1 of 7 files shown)


server.js
```diff
@@ -252,11 +250,14 @@ class Server extends BaseServer {
     })
   }
 
-  listen (...args) {
-    return BaseServer.prototype.listen.apply(this, args).catch(err => {
+  async listen (...args) {
+    try {
+      let result = await BaseServer.prototype.listen.apply(this, args)
```
ai: можно просто `return await BaseServer.prototype.listen.apply(this, args)`

ai:
We can do it separated PR. Let’s merge this PR if it passes tests. And then we will continue discussing other `Promise` parts in #55.

chafilin:
@ai Can you help, please? It fails on jsdoc step

ai:
It is recent JSDoc + Node.js 12 compatibility problem. I created the issue. Replace `node` to `"11"` in `.travis.yml`.

PR #68, opened 2019-12-26
Remove forcePromise in favor of await.

uncaught:
As requested in https://github.com/logux/server/issues/55#issuecomment-569033627

Noteworthy: The test `reports about errors during channel authorization` in `base-server.test.js` changed the order of reports.


uncaught:
Ah, sorry. I'm not familiar with forking and making pull requests. I didn't know I could not make a second commit without it ending up here.

Do I have to create a second fork for a different pull request?

ai:
In this case we can do both in the same PR

PR #70, opened 2020-03-15
Add type definitions

neruchev:


server-client.d.ts
```diff
@@ -0,0 +1,105 @@
...
+   * if (client.node.state === 'synchronized')
+   * ```
+   */
+  node: ServerNode
+
+  /**
+   * Check client subprotocol version. It uses `semver` npm package
+   * to parse requirements.
+   *
+   * @param range npm’s version requirements.
```
ai: Can you move `@param`/`@returns` below the example?

allowed-meta.d.ts
```diff
@@ -0,0 +1,17 @@
...
+ *   for (const i in meta) {
+ *     if (ALLOWED_META.includes(i)) {
+ *       filtered[i] = meta[i]
+ *     }
+ *   }
+ *   return [action, filtered]
+ * }
+ * ```
+ */
+export const ALLOWED_META: readonly ['id', 'time', 'subprotocol']
```
ai: Why do need to specify exact value in types?

base-server.d.ts
```diff
@@ -0,0 +1,704 @@
+import { Log, Store, ServerConnection, TestTime } from '@logux/core'
+import { Server as HTTPServer } from 'http'
+
+import { Context, ChannelContext, LoguxPatternParams } from './context'
+import { ServerClient } from './server-client'
+
+/**
+ * Logux meta
+ */
+export type LoguxMeta = {
```
ai: Logux Core already has `Meta`. Can we extend it as we do in Logux Client?
https://github.com/logux/client/blob/master/client/index.d.ts#L11
https://github.com/logux/client/blob/master/client/index.d.ts#L168

Also, note that `meta` can contain any other extra fields.

base-server.d.ts
```diff
@@ -0,0 +1,704 @@
...
+  reasons: string[]
+  channels?: string[]
+  clients?: string[]
+  status?: 'waiting' | 'add' | 'clean' | 'processed' | 'subscribed'
+}
+
+/**
+ * Logux base action
+ */
+export type LoguxBaseAction = {
```
ai: I think it will be better to have different `LoguxSubscribe`, `LoguxProcessed` and `LoguxUndo` types.

Right now type does not show that `type: 'logux/subscribe' can’t have `id`.

base-server.d.ts
```diff
@@ -0,0 +1,704 @@
...
+  time: number
+  subprotocol: LoguxBaseServerOptions['subprotocol']
+  server: string
+  reasons: string[]
+  channels?: string[]
+  clients?: string[]
+  status?: 'waiting' | 'add' | 'clean' | 'processed' | 'subscribed'
+}
+
+/**
```
ai: In Logux Core I do not comment utilities classes. I have comments only for classes.

base-server.d.ts
```diff
@@ -0,0 +1,704 @@
...
+ */
+export type LoguxMeta = {
+  id: string
+  time: number
+  subprotocol: LoguxBaseServerOptions['subprotocol']
+  server: string
+  reasons: string[]
+  channels?: string[]
+  clients?: string[]
+  status?: 'waiting' | 'add' | 'clean' | 'processed' | 'subscribed'
```
ai: Note that in Logux Core I hav comments for meta fields
https://github.com/logux/client/blob/master/client/index.d.ts#L22

You can copy them from https://logux.io/guide/concepts/meta/#server-meta-keys

base-server.d.ts
```diff
@@ -0,0 +1,704 @@
...
+   * SSL certificate or path to it. Path could be relative from server
+   * root. It is required in production mode, because WSS is highly
+   * recommended.
+   */
+  cert?: string
+
+  /**
+   * Function to show current server status.
+   */
+  reporter?: (event: string, payload: Object) => void
```
ai: Can we express `payload` according to `event` type like we did in Logux Client?
https://github.com/logux/client/blob/master/cross-tab-client/index.d.ts#L89-L92

base-server.d.ts
```diff
@@ -0,0 +1,704 @@
...
+
+  /**
+   * Server options.
+   *
+   * ```js
+   * console.log('Server options', server.options.subprotocol)
+   * ```
+   */
+  options: LoguxBaseServerOptions
+  reporter: LoguxBaseServerOptions['reporter'] | (() => void)
```
ai: We need a description for `reporter`

base-server.d.ts
```diff
@@ -0,0 +1,704 @@
...
+   * @param authenticator The authentication callback.
+   *
+   * ```js
+   * server.auth(async (userId, token) => {
+   *   const user = await findUserByToken(token)
+   *   return !!user && userId === user.id
+   * })
+   * ```
+   */
+  auth<Credentials = any>(authenticator: LoguxAuthenticator<Credentials>): void
```
ai: Let’s limit `credentials` to string as I did in Logux Client 
https://github.com/logux/client/blob/master/client/index.d.ts#L51

base-server.d.ts
```diff
@@ -0,0 +1,704 @@
...
+  on(event: 'fatal' | 'clientError', listener: (err: Error) => void): void
+  on<Action extends LoguxBaseAction = LoguxBaseAction>(
+    event: 'error',
+    listener: (err: Error, action: Action, meta: LoguxMeta) => void
+  ): void
+  on(
+    event: 'connected' | 'disconnected',
+    listener: (server: ServerClient) => void
+  ): void
+  on<Action extends LoguxBaseAction = LoguxBaseAction>(
```
ai: `LoguxBaseAction` works only for `logux/subscribe`, `logux/processed` and `logux/undo`.

User's action can be with any keys. The only limit is to have `type: string`.

base-server.d.ts
```diff
@@ -0,0 +1,704 @@
...
+    listener: (action: Action, meta: LoguxMeta) => void
+  ): void
+  on<Action extends LoguxBaseAction = LoguxBaseAction>(
+    event: 'processed' | 'subscribed',
+    listener: (
+      action: Action,
+      meta: LoguxMeta,
+      latencyMilliseconds: number
+    ) => void
+  ): void
```
ai: `on()` returns `() => true` function

base-server.d.ts
```diff
@@ -0,0 +1,704 @@
...
+
+/**
+ * The authentication callback.
+ *
+ * @param userId User ID.
+ * @param credentials The client credentials.
+ * @param client Client object.
+ * @returns `true` if credentials was correct
+ */
+export type LoguxAuthenticator<Credentials = any> = (
```
ai: Can you move all extra classes above the main class as I have in Logux Core?

server-client.d.ts
```diff
@@ -0,0 +1,105 @@
...
+ */
+export class ServerClient {
+  /**
+   * @param app The server.
+   * @param connection The Logux connection.
+   * @param key Client number used as `app.connected` key.
+   */
+  constructor(app: HTTPServer, connection: ServerConnection, key: number)
+
+  app: HTTPServer
```
ai: `app` should be `BaseServer`

package.json
```diff
@@ -22,6 +22,7 @@
   "license": "MIT",
   "homepage": "https://logux.io/",
   "repository": "logux/server",
+  "types": "./index.d.ts",
```
ai: Should I add it to Logux Client as well or TS can find `.d.ts` file automatically?

index.d.ts
```diff
@@ -0,0 +1,21 @@
+export { ALLOWED_META } from './allowed-meta'
+export { Server, LoguxServerOptions } from './server'
+export { ServerClient } from './server-client'
+export {
+  BaseServer,
+  LoguxActionCallbacks,
+  LoguxActionFinally,
+  LoguxAuthenticator,
```
ai: Do we need to export all used types?

allowed-meta.d.ts
neruchev: It's not obligatory. But more accurate types are more convenient.
We can replace with
```ts
export const ALLOWED_META: string[];
```

base-server.d.ts
neruchev:
```ts
server.on('connected', () => {})
```
returned `() => undefined`. Is this the expected behavior?

package.json
neruchev: TS will automatically find the file in most cases. I think this field can be deleted.

index.d.ts
neruchev: This is optional. But there are many cases where it is convenient for developers to import types in parts so that they can be used in their code. 

For example:
```ts
import { LoguxChannelAuthorizer } from '@logux/server';

const myAcessChecker:LoguxChannelAuthorizer = (ctx, action, meta) => {
  // ...
};

server.channel('foo', {
  access: myAcessChecker,
});

server.channel('bar', {
  access: myAcessChecker,
});
```

allowed-meta.d.ts
ai: I afraid that I will forget to update types after `ALLOWED_META` update

base-server.d.ts
```diff
@@ -0,0 +1,759 @@
+import { Log, Meta, Store, ServerConnection, TestTime } from '@logux/core'
+import { Server as HTTPServer } from 'http'
+
+import { Context, ChannelContext, LoguxPatternParams } from './context'
+import { ServerClient } from './server-client'
+
+export type LoguxMeta = Meta & {
```
ai: Let’s rename it to `ServerMeta` to keep name conversion (compare it with `ClientMeta` in `@logux/client`)

base-server.d.ts
```diff
@@ -0,0 +1,759 @@
...
+   * Array and client string: all clients with listed client
+   * IDs will receive the action.
+   */
+  clients?: string[]
+
+  /**
+   * Array and node string: all clients with listed node
+   * IDs will receive the action.
+   */
+  nodes?: string[]
```
ai: There is also `channel: string`, `user`, `client`, and `node` aliases

base-server.d.ts
```diff
@@ -0,0 +1,759 @@
...
+}
+
+export type LoguxUserAction = {
```
ai: You can use just `Action` from `@logux/core` here

base-server.d.ts
```diff
@@ -0,0 +1,759 @@
...
+}
+
+export type LoguxUnsubscribeAction = {
+  type: 'logux/unsubscribe'
+  channel?: string
+}
+
+export type LoguxProcessedAction = {
+  type: 'logux/processed'
+  id?: string
```
ai: We have special `ID` type in `@logux/core` (to be sure that ID was copied from the `meta.id`)

base-server.d.ts
```diff
@@ -0,0 +1,759 @@
...
+/**
+ * The authentication callback.
+ *
+ * @param userId User ID.
+ * @param credentials The client credentials.
+ * @param client Client object.
+ * @returns `true` if credentials was correct
+ */
+export type LoguxAuthenticator<Credentials = string> = (
+  userId: number | string | false,
```
ai: Let’s remove `number` at `userId` in this release. I think this feature only makes everything more complicated.

base-server.d.ts
```diff
@@ -0,0 +1,759 @@
...
+  credentials: Credentials | undefined,
```
ai: Let’s just put a string here. I think we do not need to use generic since non-string use cases are very rare (and right now Logux Client supports only strings here)

base-server.d.ts
```diff
@@ -0,0 +1,759 @@
...
+/**
+ * Check does user can do this action.
+ *
+ * @param ctx Information about node, who create this action.
+ * @param action The action data.
+ * @param meta The action metadata.
+ * @returns `true` if client are allowed to use this action.
+ */
+export type LoguxAuthorizer<
+  Action extends LoguxUserAction = LoguxUserAction
```
ai: Let’s use `A` instead of `Action`, to avoid name conflict with `Action` at Logux Core.

Also I mostly use single name for generic in short type blocks.

base-server.d.ts
```diff
@@ -0,0 +1,759 @@
...
+ * @param ctx Information about node, who create this action.
+ * @param action The action data.
+ * @param meta The action metadata.
+ * @returns Meta’s keys.
+ */
+export type LoguxResender<Action extends LoguxUserAction = LoguxUserAction> = (
+  ctx: Context,
+  action: Action,
+  meta: LoguxMeta
+) => Object | Promise<Object>
```
ai: I think we need to add `Resend` type with `{ users? string[], user?: string, … }`

base-server.d.ts
```diff
@@ -0,0 +1,759 @@
...
+  Action extends LoguxUserAction = LoguxUserAction,
+  PatternParams extends LoguxPatternParams = {}
+> = (
+  ctx: ChannelContext<PatternParams>,
+  action: Action,
+  meta: LoguxMeta
+) => LoguxChannelFilter<Action> | undefined
+
+/**
+ * Creates actions with initial state.
```
ai:
```suggestion
 * Send actions with initial state.
```

base-server.d.ts
```diff
@@ -0,0 +1,759 @@
...
+  on(event: 'fatal' | 'clientError', listener: (err: Error) => void): void
+  on<Action extends LoguxUserAction = LoguxUserAction>(
+    event: 'error',
+    listener: (err: Error, action: Action, meta: LoguxMeta) => void
+  ): void
+  on(
+    event: 'connected' | 'disconnected',
+    listener: (server: ServerClient) => void
+  ): void
+  on<Action extends LoguxAction | LoguxUserAction = LoguxUserAction>(
```
ai: All actions will be catched by `on('add', (action, meta) =>` this is why generic will not be really useful.

Developers need to use guards here:

```js
server.on('add', (action, meta) => {
  if (isUserAdd(action)) { //=> isUserAdd() changed the type with ` is UserAddAction`
    action.userId
  }
```

base-server.d.ts
```diff
@@ -0,0 +1,759 @@
...
+   *   }
+   *   async process (ctx, action, meta) {
+   *     return await phpBackend.sendHTTP(action, meta)
+   *   }
+   * })
+   * ```
+   *
+   * @param callbacks Callbacks for actions with this type.
+   */
+  otherType<Action extends LoguxUserAction = LoguxUserAction>(
```
ai: Generics is also useless for `otherType` since it will accept different types of actions

base-server.d.ts
```diff
@@ -0,0 +1,759 @@
...
+   *       server.unknownType(action, meta)
+   *     }
+   *   }
+   * })
+   * ```
+   *
+   * @param action The action with unknown type.
+   * @param meta Action’s metadata.
+   */
+  unknownType<Action extends LoguxUserAction = LoguxUserAction>(
```
ai: Why do we need generic here?

context.d.ts
```diff
@@ -0,0 +1,169 @@
...
+   *     ctx.data.user = findUser(ctx.userId)
+   *     return ctx.data.user.hasAccess(action.projectId)
+   *   }
+   *   process (ctx, action, meta) {
+   *     return ctx.data.user.rename(action.projectId, action.name)
+   *   }
+   * })
+   * ```
+   */
+  data: Object
```
ai: Do we need to add generic to `server.type` for the `data` type?

package.json
ai: Yeap, let’s delete it

ai:
Can you also add `check-dts` to the project and call it in `script.test`?

----

P. S. Thanks for working on it. I understand that we have too many notes on every code review step. It is a big project.

PR #85, opened 2020-04-15
Add platform mock for snapshot tests

andreyberezhnoy:


create-reporter/index.test.js
```diff
@@ -1,3 +1,10 @@
+let mockOS = jest.requireActual('os')
+jest
+  .spyOn(mockOS, 'platform')
```
ai, line 3: `spyOn` needs to then test that somebody called this method. Since we do not need it here, there is a better alternative:

```js
os.platform = () => 'linux'
```

create-reporter/index.test.js
```diff
@@ -66,6 +73,7 @@ function createError (name, message) {
 let originEnv = process.env.NODE_ENV
 afterEach(() => {
   process.env.NODE_ENV = originEnv
+  jest.resetModules()
```
ai, line 76: If we will not need a `spyOn`, we will not need it here too

create-reporter/index.test.js
andreyberezhnoy: `os`? not `mockOS`?

ai: Yeap, let’s call it more clear

andreyberezhnoy: If I call a variable `os`, I get an error `The module factory of `jest.mock()` is not allowed to reference any out-of-scope variables.`

ai: It is a pretty dangerous warning. I am not sure, that `mockOS` should work too.

`jest.mock` is not a function, it is a command for Babel to put its content and move it to the file replacement.

Let’s do the simple way:

```js

jest.mock('os', () => {
  return {
    platform: 'linux'
  }
})
```

andreyberezhnoy: I tried this before.
I get something like `TypeError: os.hostname is not a function` in other tests

andreyberezhnoy:
```js
let mockOS = jest.requireActual('os')
mockOS.platform = () => 'linux'
jest.mock('os', () => mockOS)
```

this works

ai:
```js
jest.mock('os', () => {
  return {
    hostname: () => 'localhost'
    platform: () => 'linux'
  }
})
```

The problem with `mockOS` that it should not work :(.

Issue #84, opened 2020-04-14
Add Server#autoloadModules


ai:
This method should be in `Server`, not `BaseServer`. It should:

1. Take `this.options.root` and load all `modules/*` files.
2. Each module will have `server => void` signature.

kabbi:
I think it's worth considering next.js filesystem-based routing experience, something like
```
.
|-- actions
|   `-- users
|       |-- add.js
|       `-- rename.js
`-- channels
    `-- users
        `-- [id].js
```
Maybe not in the core API though

ai:
This issue already fixed in `next` branch with an option to override pattern.

PR #87, opened 2020-04-16
Add autoload modules

andreyberezhnoy:
Reference #84


test/servers/autoload-modules.js
```diff
@@ -0,0 +1,15 @@
...
+let { Server } = require('../..')
+
+let app = new Server({
+  subprotocol: '1.0.0',
+  supports: '1.x',
+  root: __dirname
+})
+app.nodeId = 'server:FnXaqDxY'
+
+app.autoloadModules('modules/**/*.js')
```
ai, line 12: Let’s remove the option to check the default value

server/index.js
```diff
@@ -172,6 +173,24 @@ class Server extends BaseServer {
       return process.exit(1)
     }
   }
+
+  async autoloadModules (pattern = 'modules/**/*.js') {
```
ai, line 177: The pattern is not correct.

It will load `modules/users/index.test.js` or `modules/users/lib/utils.js`, but will not load `modules/users.js`

ai: Let’s add more file to ` test/servers/modules/`

ai:
@andreyberezhnoy you wrote realy good tests :+1:

PR #91, opened 2020-05-11
Replace bunyan with pino

AleksandrSl:


ai:
@AleksandrSl I changed branch to `next`, you will need to fix rebase conflict

human-formatter/index.js
```diff
@@ -154,34 +156,44 @@ function prettyStackTrace (c, stack, basepath) {
...
+  let c
+  if (options.color === undefined) {
+    c = chalk
+  } else {
+    c = new chalk.Instance({ level: options.color ? 3 : 0 })
   }
+  let basepath = options.basepath || process.cwd()
+  if (basepath.slice(-1) !== path.sep) basepath += path.sep
+  // Pino passes logger instance as this to prettifier constructor
+  // eslint-disable-next-line no-invalid-this
```
ai, line 176: Add override to `package.json`→`eslintConfig` to disable `no-invalid-this` for this file

human-formatter/index.js
```diff
@@ -154,34 +156,44 @@ function prettyStackTrace (c, stack, basepath) {
...
+  this.chalk = c
```
AleksandrSl, line 177: These properties are mostly added to test related code. Not the best reason, but otherwise to test basepath and chalk we should export them, or use snapshots (maybe this is the best way)

human-formatter/index.js
```diff
@@ -154,34 +156,44 @@ function prettyStackTrace (c, stack, basepath) {
...
+  // eslint-disable-next-line no-invalid-this
+  this.basepath = basepath
+
+  return function (record) {
+    // Hack to disable unwanted warning.
+    // Issue is raised to disable it more natural way
+    // https://github.com/pinojs/pino-pretty/issues/108
+    if (record.msg === PINO_FLUSH_SYNC_WARNIN_MSG) {
+      // Usually you should use undefined for logger to skip this record.
```
ai, line 186: We don’t need the second comment

human-formatter/index.js
```diff
@@ -154,34 +156,44 @@ function prettyStackTrace (c, stack, basepath) {
...
+      return ''
+    }
 
-  write (record) {
-    let c = this.chalk
     let message = [LABELS[record.level](c, record.msg)]
-
     let params = Object.keys(record)
-      .filter(i => !PARAMS_BLACKLIST[i])
+      // Can be done through redact
```
ai, line 195: Let’s avoid to add comments. I use rule “if you need a comment, change code to be more readable”

human-formatter/index.js
```diff
@@ -203,6 +215,7 @@ class HumanFormatter extends stream.Writable {
       let note = record.note
       if (typeof note === 'string') {
         note = note.replace(/`([^`]+)`/g, c.bold('$1'))
+        // TODO: Why concat with spreading?
         note = [].concat(
```
ai, line 219: It could be a bug after ES5 → ES6 migration. Try to remove it.

human-formatter/index.js
AleksandrSl: Ok, I hope issue will be fixed soon. I start a conversation to make a PR, fix is pretty simple.

human-formatter/index.js
AleksandrSl, line 195: I think I should remove comment, it can be done through redact more natural way, but will scatter logic into several places

server/index.d.ts
```diff
@@ -2,16 +2,16 @@ import BaseServer, { Logger, Reporter, BaseServerOptions } from '../base-server'
...
-   * Report process/errors to CLI in bunyan JSON or in human readable
+   * Report process/errors to CLI in pino JSON or in human readable
    * format. It can be also a function to show current server status.
    * Default is `'human'`.
    */
   reporter?: 'human' | 'json' | Reporter
 
   /**
-   * Bunyan logger with custom settings.
+   * Logger with custom settings.
```
ai, line 12: Can I ask you to add a little example here?

We didn’t have it before, but it will be nice to show how to set own logger. As a pino’s user you will do it better.

You can use some real pino logger (for instance, for Google Cloud).

AleksandrSl: I will try to add an example

human-formatter/index.js
AleksandrSl: Hm, spread operator is ES6, so I suppose this code is already ES6. 
I was just confused that such complex way had been selected. 
I'd prefer to write it with `flatMap`
`note.split('\n').flatMap(row => splitByLength(row, 80 - PADDING.length))`

server/index.d.ts
AleksandrSl: Maybe I didn't get you point about example. To discuss it I added simple example with elastic search. But in essence any logger that conforms to specified type annotation, not only pino one

ai: It looks good. You just need to fix code style in example to follow Logux rules and add somthing link:

```js
const server = new Server(
  Server.loadOptions(process, {
    …,
    logger: pino({ level: 'info' }, streamToElastic)
  })
)
```

AleksandrSl: Not so sure about whether I spot all typos

human-formatter/index.js
AleksandrSl: What about `flatMap` or it's better to stick to ES5 solution?

ai: Yeap, `flatMap` fits perfectly here.

AleksandrSl:
Nice, `flatMap` seems to be not supported in node 10, my bad

ai:
> Nice, flatMap seems to be not supported in node 10, my bad

What about `array.map().flat()`?

AleksandrSl:
From 11 version according to [MDN](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)

ai:
OK, let’s fix CI and I am ready to merge

AleksandrSl:
Added post-mortem about this case to commit message

ai:
Do you have an idea of how we can simplify logger options? Right now we have `reporter` and `logger`.

Maybe we should keep only `logger: 'human' | 'json' | PinoLogger`? I think `Reporter` API is too low level.

AleksandrSl:
> Maybe we should keep only logger: 'human' | 'json' | PinoLogger?

Sounds good, considering that `reporter: 'human' | 'json'` are ignored, if you pass custom logger.
Not sure that Reporter is low level, cause logger is wrapped in reporter and it's a single function, however it seems more complex to setup it in the right way/

What was the main point to have both reporter and logger? To allow customization of messages depending on their type?

ai:
> What was the main point to have both reporter and logger?

Mostly historical.

Do you want to remove `reporter` from the public API? (We may need it for API between `Server` and `BaseServer`, but we can keep it as non-documented private API)

AleksandrSl:
Yeah, I will try to do this

PR #93, opened 2020-05-30
Move part of reporter API to logger

AleksandrSl:
Actually 'human' and 'json' are logger types, not reporter.

At least this problem - 
> If we specify custom logger and human reporter, custom logger will also win and "human" reporter is ignored

is resolved.

This  problem
> If we specify custom logger and custom reporter, custom logger will be ignored
still exists

still persists. Maybe some warning can be issued when this situation encountered, alas, I found no examples of warnings, only errors.

### DIFF
--- a/create-reporter/index.js
+++ b/create-reporter/index.js
@@ -197,12 +197,12 @@ const REPORTERS = {
 
 function createReporter (options) {
   let logger
-  if (options.logger) {
+  if (typeof options.logger === 'object') {
     logger = options.logger
   } else {
     let stream = options.reporterStream || pino.destination()
     let prettifier = {}
-    if (options.reporter === 'human') {
+    if (options.logger === 'human') {
       let env = options.env || process.env.NODE_ENV || 'development'
       let color = env !== 'development' ? false : undefined
       prettifier = {
--- a/create-reporter/index.js
+++ b/create-reporter/index.js
@@ -195,40 +195,38 @@ const REPORTERS = {
   }
 }
 
-function createReporter (options) {
-  let logger
-  if (typeof options.logger === 'object') {
-    logger = options.logger
-  } else {
-    let stream = options.reporterStream || pino.destination()
-    let prettifier = {}
-    if (options.logger === 'human') {
-      let env = options.env || process.env.NODE_ENV || 'development'
-      let color = env !== 'development' ? false : undefined
-      prettifier = {
-        prettyPrint: {
-          basepath: options.root,
-          color
-        },
-        prettifier: humanFormatter
-      }
-    }
-    logger = pino(
-      {
-        name: 'logux-server',
-        timestamp: pino.stdTimeFunctions.isoTime,
-        ...prettifier
+function createLogger (options) {
+  let stream = options.reporterStream || pino.destination()
+  let prettifier = {}
+  if (options.logger === 'human') {
+    let env = options.env || process.env.NODE_ENV || 'development'
+    let color = env !== 'development' ? false : undefined
+    prettifier = {
+      prettyPrint: {
+        basepath: options.root,
+        color
       },
-      stream
-    )
+      prettifier: humanFormatter
+    }
   }
+  return pino(
+    {
+      name: 'logux-server',
+      timestamp: pino.stdTimeFunctions.isoTime,
+      ...prettifier
+    },
+    stream
+  )
+}
 
+function createReporter (options) {
   function reporter (type, details) {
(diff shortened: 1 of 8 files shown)


AleksandrSl:
Should I reopen https://github.com/logux/logux/issues/30 with modified description?

create-reporter/index.js
```diff
@@ -197,12 +197,12 @@ const REPORTERS = {
 
 function createReporter (options) {
```
AleksandrSl, line 198: Maybe it's also worth to split this function something like this
```
function createLogger (options) {
  if (typeof options.logger === 'object') {
    return options.logger
  } else {
    let stream = options.reporterStream || pino.destination()
    let prettifier = {}
    if (options.logger === 'human') {
      let env = options.env || process.env.NODE_ENV || 'development'
      let color = env !== 'development' ? false : undefined
      prettifier = {
        prettyPrint: {
          basepath: options.root,
          color
        },
        prettifier: humanFormatter
      }
    }
    return pino(
      {
        name: 'logux-server',
        timestamp: pino.stdTimeFunctions.isoTime,
        ...prettifier
      },
      stream
    )
  }
}

function createReporter (options) {
  function reporter (type, details) {
    let report = REPORTERS[type](details)
    let level = report.level || 'info'
    reporter.logger[level](report.details || details || {}, report.msg)
  }
  reporter.logger =
    typeof options.logger === 'object' ? options.logger : createLogger(options)
  return reporter
}
```

ai:
> Should I reopen logux/logux#30 with modified description?

Oops. Forgot about it.

Thanks for the PR. I will review it a little later (just started a day at New York timezone).

server/types.ts
```diff
@@ -1,14 +1,12 @@
...
 
 import { Server, Action, LoguxSubscribeAction } from '..'
-import pino = require('pino')
 
 let server = new Server<{ locale: string }>(
   Server.loadOptions(process, {
     subprotocol: '1.0.0',
-    reporter: 'human',
     supports: '1.x',
-    logger: pino({ name: 'logux' }),
```
ai, line 11: Why do we have this change?

create-reporter/index.js
ai: Yeap, let’s split it too

server/index.d.ts
```diff
@@ -2,11 +2,19 @@ import BaseServer, { Logger, Reporter, BaseServerOptions } from '../base-server'
 
 export type ServerOptions = BaseServerOptions & {
   /**
-   * Report process/errors to CLI in pino JSON or in human readable
-   * format. It can be also a function to show current server status.
-   * Default is `'human'`.
+   * Custom reporter for process/errors
```
ai, line 5: Let’s clarity that `reporter` is only for hacks and testing

server/index.d.ts
```diff
@@ -2,11 +2,19 @@ import BaseServer, { Logger, Reporter, BaseServerOptions } from '../base-server'
 
 export type ServerOptions = BaseServerOptions & {
```
ai, line 3: I think we need also change `./test-server/` options.

server/types.ts
AleksandrSl: Both `reporter: 'human'` and `logger` were specified here when I switched to pino, and than I didn't notice, that one of them is ignored. 
I thought that `'human'` logger should be left.
If I'm wrong I think we can use `'json'` to avoid "implementation details"

ai: We need to test that we defined `pino` types correctly for `logger` option.

AleksandrSl: Hm, seems that I don't understand purpose of this file. 
Does it contain code just to check that no errors are emitted during type check?

ai: It is a test file for `.d.ts` type definitions. Check out https://github.com/ai/check-dts

server/index.d.ts
AleksandrSl: I "analyzed" usages of `TestServer` and it seems, that we either want to provide custom reporter or use `'human'` logger with default reporter. So there is no need to add logger parameter, just remove `'human'` from reporter and make it default will suffice?

server/types.ts
AleksandrSl: Thanks!

ai:
Thanks. Good work :+1:

AleksandrSl:
Ok, but `./test-server/ ` still should be fixed

ai:
Oops. Can you send extra PR? It should be just option renaming.

PR #95, opened 2020-05-31
Fix reporter creation in test-server

AleksandrSl:
This is the most simple fix, but I don't like it. 

The only place when `'human'` is passed to `TestServer` in `logux/server` is [here](https://github.com/logux/server/blob/next/test-client/index.test.ts#L173). 
And it seems that we don't care about logs format (even test were not broken, though `'json'` logger was created actually after my reporter/logger api change), just want them to be written to provided stream.
So maybe we can remove `'human'` option and modify built-in reporter to use `reporterStream` if it's provided. (Current behaviour: when you provide `reporterStream` it's ignored if reporter is the default one)

I could try to implement my idea, the only obstacle that I don't know how `TestServer` is supposed to be used from outside.

### DIFF
--- a/test-server/index.js
+++ b/test-server/index.js
@@ -9,7 +9,7 @@ class TestServer extends BaseServer {
     if (!opts.time) {
       opts.time = new TestTime()
     }
-    if (opts.reporter === 'human') {
+    if (opts.logger === 'human') {
       opts.reporter = createReporter(opts)
     }
     opts.time.lastId += 1
--- a/test-client/index.test.ts
+++ b/test-client/index.test.ts
@@ -170,7 +170,7 @@ it('prints server log', async () => {
     write () {}
   }
   jest.spyOn(reporterStream, 'write').mockImplementation(() => {})
-  server = new TestServer({ reporter: 'human', reporterStream })
+  server = new TestServer({ logger: 'human', reporterStream })
   await server.connect('10:uuid')
   expect(reporterStream.write).toHaveBeenCalledTimes(2)
 })
--- a/test-server/index.d.ts
+++ b/test-server/index.d.ts
@@ -22,7 +22,7 @@ export type TestServerOptions = Omit<
   /**
    * Print server log to the console for debug.
    */
-  reporter?: 'human' | Reporter
+  reporter?: Reporter
 
   /**
    * Stream to be used by reporter to write log.
@@ -30,6 +30,11 @@ export type TestServerOptions = Omit<
   reporterStream?: {
     write(str: string): void
   }
+
+  /**
+   * Print logs in human readable format
+   */
+  logger?: 'human'
 }
 
 /**


test-server/index.js
```diff
@@ -10,7 +10,7 @@ class TestServer extends BaseServer {
       opts.time = new TestTime()
     }
     if (opts.reporter === 'human') {
```
ai, line 12: Why we use `reporter: 'human'` in `TestServer`, but `logger` in `BaseServer`?

ai: It will be better to rename `reporter` to `logger` in `TestServer` too

ai:
@AleksandrSl can you rename `reporter: 'human'` option in `TestServer` to `logger: 'human'` (if we pass `human` to `logger` in `BaseServer` we should do in `TestServer` too).

AleksandrSl:
More strictly we pass `logger` to `Server`, `BaseServer` doesn't have `logger` parameter. So does `TestServer`, of course I can introduce `logger` parameter to `TestServer`, if it's worth it. This will unify `Server` and `BaseServer` API in some way.

ai:
We do not need fully unifing, since I do not see any use case for JSON logger. Just `logger: 'human'` will be enouph for debuging.

AleksandrSl:
Have you meant something like this? (looks a little bit weird to me)

ai:
Yeap. Why it is weird to your?

AleksandrSl:
Cause we introduce parameter, that lead to some inconsistencies. We can pass `reporter` and `logger: 'human'` and `reporter` will be ignored

ai:
Let's remove `reporter` from `TestServer` public API

AleksandrSl:
Sounds nice, will do later today

AleksandrSl:
There is one usage of reporter [here](https://github.com/logux/server/blob/master/bind-backend-proxy/index.test.js#L256). Can be replaced with custom `reporterStream` and snapshot tests as in create-reporter tests, depends on what we wan to check here.

ai:
Snapshot tests have a lot of problems to be used outside of the “UI kit for terminal output”.

Let’s keep `reporter` as a semi-public API, which we will not promote in the docs, so only advanced users will use it.

AleksandrSl:
Didn't managed to write this in time :)
Last crazy idea: completely separate reporter and logger API. Use `reporter` as a way to compose custom report details and then use `logger` to write these reports. `reporter` won't call logger itself, so when you provide custom `reporter` you don't throw away default `logger` and other way around. 

Thanks for your time!

ai:
> Last crazy idea: completely separate reporter and logger API. Use reporter as a way to compose custom report details and then use logger to write these reports. reporter won't call logger itself, so when you provide custom reporter you don't throw away default logger and other way around.

Yeap, it is interesting.

But it should be implemnteted in both `TestServer` and `Server`.

AleksandrSl:
I will try to implement this if you don't mind

ai:
Yeap, it will be very interesting

PR #99, opened 2020-06-19
fix(base-server): cookie is empty object

euaaaio:
Cookie is always an empty object because `ws.upgradeReq` is undefined.


ai:
Thanks. Relesed in 0.9.2.

PR #100, opened 2020-06-28
Feature: Allow regex in baseServer.type

ambientlight:
**Motivation**:
Ability to specify a single processor for a set of actions. Like:

```ts
server.type(/.*TODO.*/, {
    access: function (ctx, action, meta) {
        return true;
    },
    resend: function (ctx, action, meta) {
        // Resend this action to everyone who subscribed to this user
        return { channel: "GLOBAL_TEST" };
    },
    process: function (ctx, action, meta) {
        // process
    },
    finally: function (ctx, action, meta) {
        // finally
    }
});
```

where all actions containing TODO in its name will be processed.
This small PR parses regex from types keys as long as actionName directly does not match.

### DIFF
--- a/base-server/index.js
+++ b/base-server/index.js
@@ -214,7 +214,7 @@ class BaseServer {
     this.clientIds = new Map()
     this.userIds = new Map()
     this.types = {}
-    this.regexTypes = {}
+    this.regexTypes = new Map()
     this.processing = 0
 
     this.lastClient = 0
@@ -333,22 +333,17 @@ class BaseServer {
   }
 
   type (name, callbacks) {
+    if (typeof name === 'function') name = name.toString()
+    if (this.types[name] && !(name instanceof RegExp)) {
+      throw new Error(`Action type ${name} was already defined`)
+    }
+    if (!callbacks || !callbacks.access) {
+      throw new Error(`Action type ${name} must have access callback`)
+    }
+
     if (name instanceof RegExp) {
-      if (this.regexTypes[name]) {
-        throw new Error(`Action type ${name} was already defined`)
-      }
-      if (!callbacks || !callbacks.access) {
-        throw new Error(`Action type ${name} must have access callback`)
-      }
-      this.regexTypes[name] = callbacks
+      this.regexTypes.set(name, callbacks)
     } else {
-      if (typeof name === 'function') name = name.toString()
-      if (this.types[name]) {
-        throw new Error(`Action type ${name} was already defined`)
-      }
-      if (!callbacks || !callbacks.access) {
-        throw new Error(`Action type ${name} must have access callback`)
-      }
       this.types[name] = callbacks
     }
   }
@@ -739,7 +734,7 @@ class BaseServer {
     let processor = this.types[type]
     if (processor) {
       return processor
-    } else if (Object.keys(this.regexTypes).length > 0) {
+    } else if (this.regexTypes.size > 0) {
       return this.getRegexProcessor(type) || this.otherProcessor
     } else {
       return this.otherProcessor
@@ -747,13 +742,10 @@ class BaseServer {
   }
 
   getRegexProcessor (type) {
-    let regexps = Object.keys(this.regexTypes).filter(
-      key => key.startsWith('/') && key.endsWith('/') && key.length > 1
+    let match = Array.from(this.regexTypes.keys()).find(
+      regexp => type.match(regexp) !== null
     )
-    let match = regexps.find(
-      regexp => type.match(regexp.substr(1, regexp.length - 2)) !== null
-    )
-    return match !== undefined ? this.regexTypes[match] : undefined
+    return match !== undefined ? this.regexTypes.get(match) : undefined
   }
 
   finally (processor, ctx, action, meta) {
--- a/base-server/index.js
+++ b/base-server/index.js
@@ -84,7 +84,12 @@ class BaseServer {
         if (!meta.subprotocol) {
           meta.subprotocol = this.options.subprotocol
         }
-        if (!this.options.backend && !this.types[action.type] && !isLogux) {
+        if (
+          !this.options.backend &&
+          !this.types[action.type] &&
+          !this.getRegexProcessor(action.type) &&
+          !isLogux
+        ) {
           meta.status = 'processed'
         }
       }
@@ -701,7 +706,11 @@ class BaseServer {
   }
 
   isUseless (action, meta) {
-    if (meta.status !== 'processed' || this.types[action.type]) {
+    if (
+      meta.status !== 'processed' ||
+      this.types[action.type] ||
+      this.getRegexProcessor(action.type)
+    ) {
       return false
     }
     for (let i of ['channels', 'nodes', 'clients', 'users']) {
--- a/base-server/index.js
+++ b/base-server/index.js
@@ -333,24 +333,24 @@ class BaseServer {
   }
 
   type (name, callbacks) {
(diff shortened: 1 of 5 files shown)


ai:
What is use case for this feature?

ambientlight:
@ai: I have two usecases in mind:

1. cleaner logic when we want to handle 2 different actions in the same way :

Without regex matching we would need to have slightly more boilerplate: two handlers that will call the same function inside its processor:

```ts
server.type('ADD_TODO', {
    access: function (ctx, action, meta) {
        return accessCheck(action, ctx)
    },
    process: function (ctx, action, meta) {
       return doProcess(action, ctx)  
    }
});

server.type('DELETE_TODO', {
    access: function (ctx, action, meta) {
        return accessCheck(action, ctx)
    },
    process: function (ctx, action, meta) {
       return doProcess(action, ctx)  
    }
});
```

With regex matching we now can explicitly match the actions via regex so a single processor is possible:
```ts
server.type(/.*TODO.*/, {
    access: function (ctx, action, meta) {
        return accessCheck(action, ctx)
    },
    process: function (ctx, action, meta) {
       return doProcess(action, ctx)  
    }
});
```

2. Isomorphic redux with reducers and middleware on the server-side. Em... you got the idea. Basically we drop a regex matching the beginning of the action, then each processor will do a dispatch on the server-side store.

ai:
1. We need a test to any new feature added by PR
2. It will be nice to have this feature without performance regression for the string types. Let's save RegExp callbacks to a separated list and use it only if we didn't find processor by string type.

ambientlight:
> We need a test to any new feature added by PR

@ai: yeah, saw travis, adding it now. 

> It will be nice to have this feature without performance regression for the string types. Let's save RegExp callbacks to a separated list and use it only if we didn't find processor by string type.

@ai: got it, alright, so `this.regexTypes` for the name would fit right?

ai:
> so this.regexTypes for the name would fit right?

👍

ambientlight:
@ai:
So look, maybe I am done here, not sure, have a few question here:

1. I had to define a separate method for regex type, otherwise **check-dts** will complain on the overload. What do you think here?  

2. I had to patch `isUseless` and `this.on('preadd', (action, meta) => {})` which now incur this before-mentioned overhead. What should I do here?  

3. in bind-backend-proxy, there is a reference to app.types:   `if (!app.types[action.type] && !app.getRegexProcessor(action.type)){…}`  seems to logically fit, but I don’t fully understand the logic here yet, would something like this be fine there?

ai:
> I had to define a separate method for regex type, otherwise check-dts will complain on the overload. What do you think here?  

Very strange. TypeScript should allow something like:

```ts
  type<A extends Action = AnyAction, D extends object = {}> (
    regex: RegExp,
    callbacks: ActionCallbacks<A, D, H>
  ): void

  type<A extends Action = AnyAction, D extends object = {}> (
    name: A['type'],
    callbacks: ActionCallbacks<A, D, H>
  ): void
```

What error do you have? Maybe it is related with my `name: A['type']` trick.

> I had to patch isUseless and this.on('preadd', (action, meta) => {}) which now incur this before-mentioned overhead. What should I do here?  

`isUseless` was created to prevent typo in action type. If action has no `meta.clients` or server’s processors (defined by `Server#type`) we show a warning, that seems like you had a typo.

I think you can pass a useless action through regexps to detect that there is a regexp-type to process it.

> in bind-backend-proxy, there is a reference to app.types:   if (!app.types[action.type] && !app.getRegexProcessor(action.type)){…}  seems to logically fit, but I don’t fully understand the logic here yet, would something like this be fine there?

If we have a `Server#type` processor. `!app.types[action.type] && !app.getRegexProcessor(action.type)` looks good.

ambientlight:
> What error do you have? Maybe it is related with my name: A['type'] trick.

Might be, here is exactly what I get:

Screen Shot 2020-06-30 at 10 31 54 AM

ai:
Nope. This error could be related with some issue in the new TS code.

Can you update the PR so I will see `.d.ts` change caused this error messages?

ambientlight:
@ai: you are correct, last error still remains though in travis:

```
83:31: Wrong error
273  Expected: '"bad"' is not assignable to parameter of type '"user/rename"'
274  Got: Argument of type '"bad"' is not assignable to parameter of type 'RegExp | "user/rename"'.
```

ai:
For this last error you just need to update expected error message. Look for `// THROWS` line above the error.

ambientlight:
> For this last error you just need to update expected error message. Look for // THROWS line above the error.

wow, that's cool, haven't seen such thrown error checking before.
So what do you think, is this ready?

base-server/index.js
```diff
@@ -719,11 +739,23 @@ class BaseServer {
...
       return this.otherProcessor
     }
   }
 
+  getRegexProcessor (type) {
+    let regexps = Object.keys(this.regexTypes).filter(
+      key => key.startsWith('/') && key.endsWith('/') && key.length > 1
+    )
+    let match = regexps.find(
+      regexp => type.match(regexp.substr(1, regexp.length - 2)) !== null
```
ai, line 754: We can use `this.regexTypes = Map<RegExp, callbacks>()` to reduce these hacks.

```js
for (let regexp of this.regexTypes.keys()) {
  if (type.match(regexp)) {
    return this.regexTypes.get(regexp)
  }
}
```

base-server/index.js
```diff
@@ -327,14 +333,24 @@ class BaseServer {
...
-    if (this.types[name]) {
-      throw new Error(`Action type ${name} was already defined`)
-    }
-    if (!callbacks || !callbacks.access) {
-      throw new Error(`Action type ${name} must have access callback`)
+    if (name instanceof RegExp) {
+      if (this.regexTypes[name]) {
+        throw new Error(`Action type ${name} was already defined`)
+      }
+      if (!callbacks || !callbacks.access) {
```
ai, line 340: It will be nice to avoid this check duplication. You can move this check before `if (name instanceof RegExp) {` test.

base-server/index.js
```diff
@@ -327,14 +333,24 @@ class BaseServer {
...
   type (name, callbacks) {
-    if (typeof name === 'function') name = name.toString()
-    if (this.types[name]) {
-      throw new Error(`Action type ${name} was already defined`)
-    }
-    if (!callbacks || !callbacks.access) {
-      throw new Error(`Action type ${name} must have access callback`)
+    if (name instanceof RegExp) {
+      if (this.regexTypes[name]) {
+        throw new Error(`Action type ${name} was already defined`)
```
ai, line 338: It is OK to avoid this test for RegExp, since it is an advanced API (You may find that it is hard to find a regexp duplicate when you will move to `this.regexTypes: Map<RegExp, callbacks>`).

base-server/index.js
ambientlight: haven't though of that, cool, this is much better

ambientlight:
@ai: thanks for support, comments should be addressed now.

ai:
Thanks. Will it be OK to wait for a few other features to make a bigger release?

I will not have this week and I like bigger releases because I do not like to spend cool AI’s names for very small releases.

What is your Twitter handle to mention you on the Twitter?

ambientlight:
@ai: Thanks for your instant feedback on this here, appreciated. Yeah sure, I am anyway just testing and researching logux tooling for now at https://github.com/ambientlight/logux-todomvc on my own fork.

Sure.

I don't have a twitter yet actually.

PR #101, opened 2020-06-30
Replace reporter with report event and introduce logger options

AleksandrSl:
Implements ideas from https://github.com/logux/logux/issues/40


server/index.d.ts
```diff
@@ -1,32 +1,25 @@
...
-   *   reporter: (name, details) => {
-   *     console.log('Event:', name)
-   *     console.log('Details:', JSON.stringify(details))
-   *   }
-   * })
-   * ```
+   * Logger message format.
    */
-  reporter?: Reporter
+  type?: 'human' | 'json'
```
AleksandrSl, line 7: Custom logger is excluded from here, cause it conflicts with stream option. (It will not respect it)

server/index.d.ts
```diff
@@ -1,32 +1,25 @@
...
 
   /**
-   * Stream to be used by reporter to write log.
+   * Stream to be used by logger to write log.
+   * todo Note about flushSync
```
AleksandrSl, line 11: Streams should have `flushSync` method to guarantee final log writes. Do we need to document this?

base-server/index.js
```diff
@@ -27,7 +27,12 @@ function optionError (msg) {
 class BaseServer {
   constructor (opts = {}) {
     this.options = opts
-    this.reporter = this.options.reporter || function () {}
+    this.reporter = function (type, details) {
+      this.emitter.emit('report', type, details)
```
AleksandrSl, line 31: I chose to always emit event, cause there are cases when we want to listen for reports, but don't want them printed to logs

ai: Why do we still need `this.reporter` and `opts.reporter`?

AleksandrSl: Fair remark

AleksandrSl: We still use reporter inside app, e.g. https://github.com/logux/server/pull/101/files#diff-2a339baf3d65ca89f18b2d3a81f4634fR105 and in server

ai: I can't open a link from the phone. What is the file name and line?

server/index.d.ts
ai: Yeap. And we may want to define it in types.

base-server/index.js
AleksandrSl: base-server/index.js 105 line

server/index.d.ts
ai: We will need a way to add, for instance, [Google Cloud pino driver](https://github.com/ovhemert/pino-stackdriver/blob/master/docs/API.md).

Do we have the way without using `server.on('report'`?

base-server/index.js
ai: Why we can’t replace it to:

```js
this.emitter.emit('report', type, details)
```

And all `options.reporters` usage can be replaced to `server.on('report'` (if we used it for internal code or dirty hacks).

ai:
Sorry for the long delay. I had a rough period with PostCSS 8 release and now I am back on Logux.

I rebased the branch, fixed CI, and answered old comments.

AleksandrSl:
Thanks! I will reimagine this PR during weekends

server/index.d.ts
AleksandrSl: We can pass a custom logger, as described in docs, since we still have `logger?: Logger | LoggerOptions`
   ```js
    const pino = require('pino')
    const pinoElastic = require('pino-elasticsearch')
   
    const streamToElastic = pinoElastic({
      consistency: 'one',
      node: 'http://localhost:9200',
      ecs: true
    })
   
    const server = new Server(
      Server.loadOptions(process, {
        …,
        logger: pino({ level: 'info' }, streamToElastic)
      })
    )
```

`LoggerOptions` allows us to customize built-in logger.
Besides most of pino custom drivers differs only in stream as in example above, so we can do:

```js
const pinoElastic = require('pino-elasticsearch')
   
const streamToElastic = pinoElastic({
  consistency: 'one',
  node: 'http://localhost:9200',
  ecs: true
})
   
const server = new Server(
  Server.loadOptions(process, {
    …,
    logger: { stream: streamToElastic }
  })
)
```

with google it should be:
```js
const stackdriver = require('pino-stackdriver')

// create the stackdriver destination stream
const credentials = '/credentials.json'
const projectId = 'my-project'
const writeStream = stackdriver.createWriteStream({ credentials, projectId })

const server = new Server(
  Server.loadOptions(process, {
    …,
    logger: { stream: writeStream }
  })
)
```

base-server/index.js
AleksandrSl: Sounds reasonable

ai:
Looks amazing 😍

It was a great idea

PR #110, opened 2021-02-22
Replace yargs with custom args parser

AleksandrSl:
Aims:
 * Drop yargs
 * Add colors to simplify help perception
 * Unify handling of CLI, ENV, and JS arguments

Colored help looks like this:
Screenshot from 2021-02-22 10-54-26

Improvements:
* Commands like `node ./test/servers/options.js --por 4545` (wrong argument spelling) and `node ./test/servers/options.js --port 4535l` (wrong type of value) are not allowed now. With yargs they are just ignored, and the server starts with defaults

Decisions:
* We have 3 API. CLI and ENV require parsing and validation, JS requires only validation. Hence, validation is separated from parsing.
* All arguments that could be passed via CLI could be passed via ENV. This coupling seems adequate and simplifies configuration.
* To avoid triplication of API arguments, they are specified in one place.
* Arg parser and validation is project independent

Questions:
* Why `fileUrl` has special treatment in `loadOptions`
* Why we check that no unknown options are given after server construction? Can we do this check before?

If the overall design looks reasonable, I will, of course, add tests and cleanup code.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1,9 +1,10 @@
 import { join, relative } from 'path'
 import globby from 'globby'
+import { bgRed, black } from 'colorette'
 
 import { createReporter } from '../create-reporter/index.js'
 import { BaseServer } from '../base-server/index.js'
-import { loadOptions, oneOf } from '../options-loader/index.js'
+import { loadOptions, oneOf, number } from '../options-loader/index.js'
 
 let cliOptionsSpec = {
   options: {
@@ -14,7 +15,7 @@ let cliOptionsSpec = {
     port: {
       alias: 'p',
       description: 'Port to bind server',
-      parse: it => Number.parseInt(it, 10)
+      parse: number
     },
     key: {
       description: 'Path to SSL key'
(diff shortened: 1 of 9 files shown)


server/index.js
```diff
@@ -1,144 +1,114 @@
...
-    if (defaults.root) {
-      dotenv.config({ path: join(defaults.root, '.env') })
-    } else {
-      dotenv.config()
-    }
-    let argv = yargs.parse(process.argv)
-    let opts = {}
-
-    for (let name of AVAILABLE_OPTIONS) {
-      if (name !== 'fileUrl') {
```
AleksandrSl, line 127: Why fileUrl has special treatment?

server/index.js
```diff
@@ -167,20 +137,6 @@ export class Server extends BaseServer {
       }
     })
 
-    for (let name in opts) {
```
AleksandrSl, line 170: Why we check that no unknown options are given after server construction? Can we do this check before?

We will not be able to use a reporter to log an error, but do we really need a formatted error when a server cannot be started?

server/index.js
ai: `fileUrl` has no CLI argument or environment variable, it is a constructor option only for the case of ESM application (to avoid conversation `import.meta.url` to `__dirname`).

server/index.js
ai: > We will not be able to use a reporter to log an error, but do we really need a formatted error when a server cannot be started?

Nope. We can even remove it. Let’s users use TypeScript if they want extra checks.

server/index.js
AleksandrSl: Ok, in my design it's in JS api only

AleksandrSl:
Ok, I decided to drop validation of arguments passed through JS API.
Reasons:
* Complicates code
* Could be done through TS
* We cannot validate options totally. e.g., we can pass custom loggers, checking that the implement required API is a frenzy

I will also remove the ability to use CLI argument aliases like `--l` and `--p`. A single dash (`-l`) is a common practice.

ai:
> I decided to drop validation of arguments passed through JS API.

Yeap, let’s remove it. I plan to keep constructor API as a private API in future versions of Logux Server.

ai:
> I will also remove the ability to use CLI argument aliases like --l and --p. A single dash (-l) is a common practice.

👍

PR #112, opened 2021-04-15
Introduce channelCallbacks.unsubscribe hook

erictheswift:
Implements https://github.com/logux/logux/issues/58

### DIFF
--- a/base-server/index.d.ts
+++ b/base-server/index.d.ts
@@ -360,6 +360,26 @@ interface ChannelFinally<
   (ctx: ChannelContext<D, P, H>, action: A, meta: ServerMeta): void
 }
 
+/**
+ * Callback which will be called on listener unsubscribe
+ * (with explicit intent or because of disconnect)
+ *
+ * @param ctx Information about node, who create this action.
+ * @param action The action data.
+ * @param meta The action metadata.
+ */
+interface ChannelUnsubscribe<
+  D extends object,
+  P extends object | string[],
+  H extends object
+> {
+  (
+    ctx: ChannelContext<D, P, H>,
+    action: LoguxUnsubscribeAction,
+    meta: ServerMeta
+  ): void
+}
+
 type ActionCallbacks<A extends Action, D extends object, H extends object> =
   | {
       access: Authorizer<A, D, H>
@@ -384,6 +404,7 @@ type ChannelCallbacks<
       filter?: FilterCreator<A, D, P, H>
       load?: ChannelLoader<A, D, P, H>
       finally?: ChannelFinally<A, D, P, H>
+      unsubscribe?: ChannelUnsubscribe<D, P, H>
     }
   | {
       accessAndLoad: ChannelLoader<A, D, P, H>
@@ -648,7 +669,10 @@ export class BaseServer<
    */
   subscribers: {
     [channel: string]: {
-      [nodeId: string]: ChannelFilter<{}> | true
+      [nodeId: string]: {
+        filter: ChannelFilter<{}> | true
+        unsubscribe?: ChannelUnsubscribe<any, any, any>
+      }
     }
   }
 
--- a/base-server/index.js
+++ b/base-server/index.js
@@ -784,16 +784,16 @@ export class BaseServer {
     if (!match) this.wrongChannel(action, meta)
   }
 
-  unsubscribe(action, meta) {
-    let nodeId = meta.id.split(' ')[1]
+  performUnsubscribe(clientNodeId, action, meta) {
     if (this.subscribers[action.channel]) {
-      let subscriber = this.subscribers[action.channel][nodeId]
+      let subscriber = this.subscribers[action.channel][clientNodeId]
       if (subscriber) {
         if (subscriber.unsubscribe) {
           let ctx = this.createContext(action, meta)
           subscriber.unsubscribe(ctx, action, meta)
+          this.contexts.delete(action)
         }
-        delete this.subscribers[action.channel][nodeId]
+        delete this.subscribers[action.channel][clientNodeId]
         if (Object.keys(this.subscribers[action.channel]).length === 0) {
           delete this.subscribers[action.channel]
         }
@@ -804,6 +804,11 @@ export class BaseServer {
       actionId: meta.id,
       channel: action.channel
     })
+  }
+
+  unsubscribe(action, meta) {
+    let clientNodeId = meta.id.split(' ')[1]
+    this.performUnsubscribe(clientNodeId, action, meta)
   }
 
   unsubscribeAction(action, meta) {
--- a/base-server/index.js
+++ b/base-server/index.js
@@ -539,7 +539,7 @@ export class BaseServer {
       if (!this.subscribers[channel]) {
         this.subscribers[channel] = {}
       }
-      this.subscribers[channel][nodeId] = true
+      this.subscribers[channel][nodeId] = { filter: true }
       this.log.add({ type: 'logux/subscribed', channel }, { nodes: [nodeId] })
     }
   }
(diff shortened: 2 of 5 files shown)


server-client/index.js
```diff
@@ -80,10 +80,13 @@ class ServerClient {
...
-      for (let i in this.app.subscribers) {
-        delete this.app.subscribers[i][this.nodeId]
-        if (Object.keys(this.app.subscribers[i]).length === 0) {
-          delete this.app.subscribers[i]
+      let lastDestroyIndex = 0
+      for (let channel in this.app.subscribers) {
+        if (this.app.subscribers[channel][this.nodeId]) {
+          this.app.log.add(
+            { type: 'logux/unsubscribe', channel },
+            { id: `destroy${++lastDestroyIndex} ${this.nodeId}` }
```
ai, line 88: It is not valid [Logux ID](https://logux.io/guide/concepts/meta/#id-and-time).

It is better to avoid generating ID from another node since it could be not unique.

base-server/index.d.ts
```diff
@@ -610,7 +631,10 @@ export default class BaseServer<
    */
   subscribers: {
     [channel: string]: {
-      [nodeId: string]: ChannelFilter<{}> | true
+      [nodeId: string]: {
+        filter: ChannelFilter<{}> | true
+        channelUnsubscribeCallback?: ChannelUnsubscribe<any, any, any>
```
ai, line 636:
```suggestion
        unsubscribe?: ChannelUnsubscribe<any, any, any>
```

Let’s make it shorter to keep code style consistency/

base-server/index.js
```diff
@@ -607,7 +610,10 @@ class BaseServer {
             this.subscribers[action.channel] = {}
             this.emitter.emit('subscribing', action, meta)
           }
-          this.subscribers[action.channel][ctx.nodeId] = filter
+          this.subscribers[action.channel][ctx.nodeId] = {
+            filter,
+            channelUnsubscribeCallback: channel.unsubscribe
```
ai, line 615:
```suggestion
            unsubscribe: channel.unsubscribe
```

base-server/index.js
```diff
@@ -648,9 +654,16 @@ class BaseServer {
 
     let nodeId = meta.id.split(' ')[1]
     if (this.subscribers[action.channel]) {
-      delete this.subscribers[action.channel][nodeId]
-      if (Object.keys(this.subscribers[action.channel]).length === 0) {
-        delete this.subscribers[action.channel]
+      let subscriber = this.subscribers[action.channel][nodeId]
+      if (subscriber) {
+        if (subscriber.channelUnsubscribeCallback) {
```
ai, line 659:
```suggestion
        if (subscriber.unsubscribe) {
```

base-server/index.test.ts
```diff
@@ -927,9 +928,9 @@ it('subscribes clients', async () => {
   })
   expect(test.reports[3][1].meta.clients).toEqual(['10:a'])
   expect(test.reports[3][1].meta.status).toEqual('processed')
-  expect(test.app.subscribers).toEqual({
+  expect(test.app.subscribers).toMatchObject({
```
ai, line 931: `toMatchObject` is required to templates like `expect.stringMatching(/white|yellow/)`

Let’s use `toEqual` if we do not use templates.

server-client/index.js
ai: It means that logic will be more complicated. Seems like we will need to call the `unsubscribe` callback manually in the `disconnect` callback.

base-server/index.d.ts
erictheswift: Done

server-client/index.js
erictheswift: The purpose of faking the id is to make this action be processed as if it was emitted by a client (nodeId is extracted from id right now).

What would you suggest to solve this? Replacing `destroy{N}` with server local time?

erictheswift: So your suggestion is to get rid of emitting 'logux/unsubscribe' in case of disconnect and duplicate this logic in ServerClient?

ai: > So your suggestion is to get rid of emitting 'logux/unsubscribe' in case of disconnect

You may want to create `logux/unsubscribe` action to pass it to `unsubscribe` handler.

> duplicate this logic in ServerClient

Yeap. But of course you can avoid duplication by moving this logic to some private method in `BaseServer`.

server-client/index.test.ts
```diff
@@ -220,12 +220,19 @@ it('reports about connection', () => {
   expect(fired).toEqual(['1'])
 })
 
-it('removes itself on destroy', async () => {
+it('emits logux/unsubscribe action for every subscription channel on destroy', async () => {
   let test = createReporter()
-  let fired: string[] = []
+  let clientKeysDisconnectEventWasFiredFor: string[] = []
```
ai, line 225: Let’s make it shorter. For instance, `disconnectedKeys`

erictheswift: Done

base-server/index.test.ts
erictheswift: Mainly yes but also may be useful to keep only meaningful fields enumerated in test.

Fixed for subscribers (really redundant since I stopped using null-safe `noop` function as a `subscriber.unsubscribe` default)

server-client/index.js
erictheswift: Done

server-client/index.js
```diff
@@ -82,10 +82,17 @@ export class ServerClient {
...
-          delete this.app.subscribers[i]
+      let clientNodeLog = new Log({
+        nodeId: this.nodeId,
+        store: new MemoryStore()
+      })
+      for (let channel in this.app.subscribers) {
+        let subscriber = this.app.subscribers[channel][this.nodeId]
+        if (subscriber) {
+          let action = { type: 'logux/unsubscribe', channel }
+          let actionId = clientNodeLog.generateId()
```
ai, line 93: It could generate an already used ID. You need to generate ID by `server.log.generateId()`, which leads to calling unsubscribe callback manually.

erictheswift: Makes sense, done

server-client/index.js
ai: Looks great. Just fix the test and we are ready for merge.

erictheswift: Done :)

erictheswift: btw is that only my local macOS setup issue or `reports EADDRINUSE error` test snapshot is outdated?
```
  ● reports EADDRINUSE error

    expect(received).toMatchSnapshot()

    Snapshot name: `reports EADDRINUSE error 1`

    - Snapshot  - 1
    + Received  + 1

    - {"level":60,"time":"1970-01-01T00:00:00.000Z","pid":21384,"hostname":"localhost","name":"test","err":{},"note":"Another Logux server or other app already running on this port. Probably you haven’t stopped server from other project or previous version of this server was not killed.\n\n$ su - root\n# netstat -nlp | grep 31337\nProto   Local Address   State    PID/Program name\ntcp     0.0.0.0:31337    LISTEN   `777`/node\n# sudo kill -9 `777`","msg":"Port `31337` already in use"}
    + {"level":60,"time":"1970-01-01T00:00:00.000Z","pid":21384,"hostname":"localhost","name":"test","err":{},"note":"Another Logux server or other app already running on this port. Probably you haven’t stopped server from other project or previous version of this server was not killed.\n\n$ sudo lsof -i 31337\n$ sudo kill -9 `<processid>`","msg":"Port `31337` already in use"}
      ↵
```

```
  ● reports EADDRINUSE error

    expect(received).toMatchSnapshot()

    Snapshot name: `reports EADDRINUSE error 2`

    - Snapshot  - 5
    + Received  + 2

       FATAL  Port 31337 already in use at 1970-01-01 00:00:00
              Another Logux server or other app already running on this port.
              Probably you haven’t stopped server from other project or previous
              version of this server was not killed.

    -         $ su - root
    -         # netstat -nlp | grep 31337
    -         Proto   Local Address   State    PID/Program name
    -         tcp     0.0.0.0:31337    LISTEN   777/node
    +         $ sudo lsof -i 31337
    -         # sudo kill -9 777
    +         $ sudo kill -9 <processid>
```

erictheswift: (that is my lame reason of missing failing test in plethora of other failing test output:)

ai:
Thanks. Great work. 

I will implement a few things on this week and will start docs to make a release.

PR #113, opened 2021-04-23
Preserve subscription context { data, headers, params } for channel unsubscribe callback

erictheswift:
FYI the reason of failing build (.aValue typecheck failure) is present in current `next` branch

### DIFF
--- a/base-server/index.test.ts
+++ b/base-server/index.test.ts
@@ -1177,7 +1177,7 @@ it('calls unsubscribe() channel callback with logux/unsubscribe', async () => {
   let client: any = {
     node: {
       remoteSubprotocol: '0.0.0',
-      remoteHeaders: { preservedSubscribedClientHeaders: true },
+      remoteHeaders: { preservedClientHeaders: true },
       onAdd: () => false
     }
   }
@@ -1190,13 +1190,16 @@ it('calls unsubscribe() channel callback with logux/unsubscribe', async () => {
     meta.reasons.push('test')
   })
   let unsubscribeCallback = jest.fn()
-  test.app.channel<{ id: string }>('user/:id', {
-    access(ctx) {
-      ;(ctx.data as any).preservedInitialData = true
-      return true
-    },
-    unsubscribe: unsubscribeCallback
-  })
+  test.app.channel<{ id: string }, { preservedInitialData?: boolean }>(
+    'user/:id',
+    {
+      access(ctx) {
+        ctx.data.preservedInitialData = true
+        return true
+      },
+      unsubscribe: unsubscribeCallback
+    }
+  )
 
   await test.app.log.add(
     { type: 'logux/subscribe', channel: 'user/10' },
@@ -1223,7 +1226,7 @@ it('calls unsubscribe() channel callback with logux/unsubscribe', async () => {
       userId,
       clientId,
       data: { preservedInitialData: true },
-      headers: { preservedSubscribedClientHeaders: true },
+      headers: { preservedClientHeaders: true },
       params: { id: '10' },
       subprotocol: '0.0.0'
     }),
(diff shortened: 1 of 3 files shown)


erictheswift:
@ai I think it's quite important to make a note in next documentation that
`ctx.userId`|`ctx.nodeId` in case of disconnect-related unsubscribe is `server` (as we've decided, this action is emitted by server for correctness), thus real subscriber user id should be sourced from stored `ctx.data`/`ctx.headers`

ai:
Nope, `ctx.userId | ctx.nodeId` should be not be taken from that fake unsubscribe action.

erictheswift:
@ai that's what I'm talking about — that's not explicit and should be documented

ai:
> that's not explicit and should be documented

Nobody read the docs. API should not send `server` as `ctx.userId`.

erictheswift:
Yeah that would be great. I think `nodeId` then also should be inherited from subscription user context, right?

ai:
You can re-use the whole context without creating a new one.

erictheswift:
Ok, we're sticking to the context that was used at the moment of subscription.
What about `serverMeta`?
Should it be related to initial subscribe action or unsubscribe action (with `server` node in id sometimes)?

ai:
> What about serverMeta?

Nope, it is OK to use `server` in `meta` since it is rare to look for user there. `meta` is mostly for time.

erictheswift:
👌

base-server/index.test.ts
```diff
@@ -1175,17 +1175,26 @@ it('loads initial actions during subscription', async () => {
...
+  let clientId = '10:uuid'
+  let userId = '10'
   test.app.nodeIds.set(nodeId, client)
-  test.app.clientIds.set('10:uuid', client)
+  test.app.clientIds.set(clientId, client)
   test.app.on('preadd', (action, meta) => {
     meta.reasons.push('test')
   })
   let unsubscribeCallback = jest.fn()
   test.app.channel<{ id: string }>('user/:id', {
```
ai, line 1193:
```suggestion
  test.app.channel<{ id: string }, { preservedInitialData?: boolean }>('user/:id', {
```

Will allow to avoid `any`.

base-server/index.test.ts
```diff
@@ -1175,17 +1175,26 @@ it('loads initial actions during subscription', async () => {
 it('calls unsubscribe() channel callback with logux/unsubscribe', async () => {
   let test = createReporter({})
   let client: any = {
-    node: { remoteSubprotocol: '0.0.0', onAdd: () => false }
+    node: {
+      remoteSubprotocol: '0.0.0',
+      remoteHeaders: { preservedSubscribedClientHeaders: true },
```
ai, line 1180:
```suggestion
      remoteHeaders: { preservedClientHeaders: true },
```

Let’s make them shorter

erictheswift:
👌

base-server/index.test.ts
```diff
@@ -1175,19 +1175,31 @@ it('loads initial actions during subscription', async () => {
...
+  test.app.clientIds.set(clientId, client)
   test.app.on('preadd', (action, meta) => {
     meta.reasons.push('test')
   })
   let unsubscribeCallback = jest.fn()
-  test.app.channel<{ id: string }>('user/:id', {
-    access: () => true,
-    unsubscribe: unsubscribeCallback
-  })
+  test.app.channel<{ id: string }, { preservedInitialData?: boolean }>(
```
ai, line 1193:
```suggestion
  test.app.channel<{ id: string }, { preservedData?: boolean }>(
```

It will allow to write a arguments in one line

ai:
Tests didn’t passed

erictheswift:
This PR does not contribute anything to failing tests.
>    base-server/index.test.ts:1368:29 - error TS2339: Property 'aValue' does not exist on type 'Action'.

is failing in `next` branch as I've mentioned in PR description:

> FYI the reason of failing build (.aValue typecheck failure) is present in current next branch

PR #115, opened 2021-05-20
Allow any character save for `/` in channel name

AleksandrSl:
Initial request - allow base64 encoded names. `=` was prohibited. We don't seem to use names in any context where a name should be restricted to some symbols


ai:
Thanks. Released in 0.10.1.

PR #118, opened 2021-09-03
Fix processing of server actions when `accessAndProcess` callback is registered

AleksandrSl:
Thought about adding flag directly to `normalizeTypeCallbacks`, like
```
if (callbacks && callbacks.accessAndProcess) {
  callbacks.access = (...args) => {
    callbacks.skipProcess = true
    return wasNot403(async () => {
      await callbacks.accessAndProcess(...args)
    })
  }
  callbacks.process = async (...args) => {
    if (callbacks.skipProcess) {
      callbacks.skipProcess = false
      return
    }
    await callbacks.accessAndProcess(...args)
  }
}
```
But this state should be in the action not in the listener. e.g. some asynchronous magic and two actions call their access methods -> only one processed will be stopped.

Then I tried to set status: 'processed' but this messed up the whole action processing.

Thus, I decided to put the flag to action meta via Symbol not to conflict with user props.

But @ai said that it could ruin for example meta serialization to database, and we have context that is created per action, so it could be used.

### DIFF
--- a/base-server/index.js
+++ b/base-server/index.js
@@ -13,6 +13,7 @@ import { createHttpServer } from '../create-http-server/index.js'
 import { ServerClient } from '../server-client/index.js'
 import { Context } from '../context/index.js'
 
+const SKIP_PROCESS = Symbol('skipProcess')
 const RESEND_META = ['channels', 'users', 'clients', 'nodes']
 
 function optionError(msg) {
@@ -54,10 +55,14 @@ export class LoguxNotFoundError extends Error {
 
 function normalizeTypeCallbacks(name, callbacks) {
   if (callbacks && callbacks.accessAndProcess) {
-    callbacks.access = (...args) => {
+    callbacks.access = (ctx, ...args) => {
       return wasNot403(async () => {
-        await callbacks.accessAndProcess(...args)
+        await callbacks.accessAndProcess(ctx, ...args)
+        ctx[SKIP_PROCESS] = true
       })
+    }
+    callbacks.process = async (ctx, ...args) => {
+      if (!ctx[SKIP_PROCESS]) await callbacks.accessAndProcess(ctx, ...args)
     }
   }
   if (!callbacks || !callbacks.access) {
--- a/server-client/index.test.ts
+++ b/server-client/index.test.ts
@@ -1769,6 +1769,49 @@ it('has shortcut to access and process in one callback', async () => {
   ])
 })
 
+it('process action exactly once with accessAndProcess callback', async () => {
+  let app = createServer()
+  app.log.keepActions()
+
+  app.type('FOO', {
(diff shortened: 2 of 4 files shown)


server-client/index.test.ts
```diff
@@ -1769,6 +1769,49 @@ it('has shortcut to access and process in one callback', async () => {
   ])
 })
 
+it('process actoin exactly once with accessAndProcess callback', async () => {
```
AleksandrSl, line 1772:
```suggestion
it('process action exactly once with accessAndProcess callback', async () => {
```

base-server/index.js
```diff
@@ -54,11 +55,18 @@ export class LoguxNotFoundError extends Error {
...
-    callbacks.access = (...args) => {
+    callbacks.access = (ctx, action, meta) => {
       return wasNot403(async () => {
-        await callbacks.accessAndProcess(...args)
+        await callbacks.accessAndProcess(ctx, action, meta)
+        meta[SKIP_PROCESS] = true
       })
     }
+    callbacks.process = async (ctx, action, meta) => {
+      if (meta[SKIP_PROCESS]) {
```
ai, line 65: Nope. This will not work with different log storages (for instance, somebody may want to store log in database).

I recommend to use `ctx` since we re-use `ctx` between `access` and `process` calls.

AleksandrSl: So we could just put the same flag in `ctx`?

ai: Yeap

base-server/index.js
```diff
@@ -54,11 +55,16 @@ export class LoguxNotFoundError extends Error {
...
+    callbacks.access = (ctx, action, meta) => {
       return wasNot403(async () => {
-        await callbacks.accessAndProcess(...args)
+        await callbacks.accessAndProcess(ctx, action, meta)
+        ctx[SKIP_PROCESS] = true
       })
     }
+    callbacks.process = async (ctx, action, meta) => {
+      if (!ctx[SKIP_PROCESS])
+        {await callbacks.accessAndProcess(ctx, action, meta)}
```
AleksandrSl, line 66: Seems it was done by prettier

base-server/index.js
```diff
@@ -54,11 +55,16 @@ export class LoguxNotFoundError extends Error {
 
 function normalizeTypeCallbacks(name, callbacks) {
   if (callbacks && callbacks.accessAndProcess) {
-    callbacks.access = (...args) => {
+    callbacks.access = (ctx, action, meta) => {
```
AleksandrSl, line 58: We also can do
```suggestion
    callbacks.access = (ctx, ...args) => {
```

saves a little space

ai:
Released in 0.10.2.

PR #121, opened 2021-10-07
feat(base-server): async channel filter

euaaaio:
Useful for:

```ts
filter: (ctx, action) => {
  return (resentCtx, resentAction) => {
    if (deletedTask.match(resentAction)) {
      let task = await getTask(resentAction.id)
      return task.userId === ctx.userId
    }
  }
}
```


ai:
I will release it today evening

PR #125, opened 2022-03-29
Draft: Migrate from pino 6 to pino 7

Zn250g:
Noticed issues:
  - some spnapshot tests are not stable. In one out of 10 cases, some snapshot tests (e.g., 'destroy.js') change the order of output lines


create-reporter/index.js
```diff
@@ -1,7 +1,9 @@
 import pino from 'pino'
 import os from 'os'
+import path from "path"
+import tty from 'tty'
 
-import { humanFormatter } from '../human-formatter/index.js'
+export const PATH_TO_PRETTIFYING_PINO_TRANSPORT = '../human-formatter/index.js';
```
ai, line 6:
```suggestion
export const PATH_TO_PRETTIFYING_PINO_TRANSPORT = '../human-formatter/index.js'
```

ai: Do we need to use `path.join(dir)`? Relative paths are bad to export.

create-reporter/index.js
```diff
@@ -197,26 +199,50 @@ const REPORTERS = {
   }
 }
 
+// https://github.com/alexeyraspopov/picocolors/blob/147fbdbd5aea8e49d246fd5faee5e1c4f97d560a/picocolors.js#L3
```
ai, line 202: Why we can’t use `picocolors`?

create-reporter/index.js
```diff
@@ -1,7 +1,14 @@
 import pino from 'pino'
 import os from 'os'
+import path from 'path'
```
ai, line 3: Can we move these import to `import { join, dirname } from 'path'` and `import { fileURLToPath } from 'url'` to keep consistency across the project?

Zn250g: Yes! Of course.

Sorry for not responding to the previous comments. I'm trying to make things work consistently and as expected. Lots of new things for me.

Thank you so much for your review.

Zn250g:
Hi

It is quite possible that [the tests do not pass](https://github.com/logux/server/runs/5769073818?check_suite_focus=true#step:6:836) because of the deadlock [here](https://github.com/pinojs/thread-stream/blob/v0.15.1/index.js#L480).

Working on it.

PR #131, opened 2022-06-29
feat(base-server): global filter

euaaaio:


base-server/index.d.ts
```diff
@@ -1082,6 +1082,30 @@ export class BaseServer<
     callbacks: ChannelCallbacks<LoguxSubscribeAction, Data, [string], Headers>
   ): void
 
+  /**
+   * Define the creator of filter for all resending actions.
+   *
+   * ```js
+   * server.filter((ctx, action, meta) => {
```
ai, line 1089: How they will work together with `filter` in `Server#channel`?

euaaaio: This works the same way, but all the filters are in the one array. First the global filters, then the channel filter.

euaaaio: Also, global filters are used on manually subscribing from the server side.

PR #134, opened 2022-09-03
addSyncMapFilter/initial: pass correct since value

ngalaiko:
according to the typings, initial() should receive
`since` with type `number | undefined` as it's 3rd
parameter, but it actually receives a value of the
`action.since` which type is
`{ id: ID, time: number } | undefined`.


ai:
Thanks. Released in 0.12.1.

PR #135, opened 2022-09-05
addSyncMap/load: pass correct since value

ngalaiko:
similar to https://github.com/logux/server/pull/134, type mismatch


ai:
Thanks. Released in 0.12.2.

PR #140, opened 2022-12-12
Add action queues

muddv:
Added queues as per [this issue](https://github.com/logux/logux/issues/54).
To process actions through queues, developer should add `true` as the last argument, when declaring `types` for actions in server modules; if this is omitted, actions will be processed normally. When a client connects, each gets a separate queue for their actions.

Some notes:
- My solution does not touch `core` at all: `ServerClient.filter` returns `false` to `syncMessage` in `core` if action should be processed in the queue. This allows not to check the queue in the core itself. I did not notice this breaking anything, but I may be missing something.
- Right now, there is no ability to create separate queues, all the action types will be placed in the same queue, if needed, I can modify my solution to support this functionality.
- I have not written tests for new functionality yet, but if you do not see any major problems with my implementation right away, I can write them to bring the test coverage back up to 100 before closing the pr.
- I made a [very basic test app](https://github.com/muddv/logux-queues-test), so you can see how the new functionality works there.

### DIFF
--- a/base-server/index.d.ts
+++ b/base-server/index.d.ts
@@ -689,15 +689,16 @@ export class BaseServer<
   connected: Map<string, ServerClient>
 
   /**
-   * Queue channels.
-   * Queues for processing different actions.
-   * Default queue has 'main' as a key, other queues can be
-   * created for different channels.
+   * Types of queues.
+   * 
+   * Queues for processing different actions in parallel.
+   * By default every action is processed in 'main' queue.
    */
-  queueChannels: Map<string, queueChannel>
+  queueTypes: Map<string, queueChannel>
 
   /**
-   * Queues by client ID.
+   * Queues by client ID and queue type.
+   * 
    * each queue processes actions from one client.
    */
   queues: Map<string, Queue>
--- a/base-server/index.d.ts
+++ b/base-server/index.d.ts
@@ -19,7 +19,6 @@ import { LogFn } from 'pino'
 
 import { Context, ChannelContext } from '../context/index.js'
 import { ServerClient } from '../server-client/index.js'
-import { Queue } from '../queue/index.js'
 
 export interface ServerMeta extends Meta {
   /**
@@ -689,21 +688,6 @@ export class BaseServer<
   connected: Map<string, ServerClient>
 
   /**
-   * Types of queues.
-   *
-   * Queues for processing different actions in parallel.
-   * By default every action is processed in 'main' queue.
-   */
-  queueTypes: Map<string, string>
-
-  /**
-   * Queues by client ID and queue type.
-   *
-   * each queue processes actions from one client.
-   */
-  queues: Map<string, Queue>
-
-  /**
    * Connected client by client ID.
    *
    * Do not rely on this data, when you have multiple Logux servers.
@@ -1069,7 +1053,7 @@ export class BaseServer<
   /**
    * @param pattern Regular expression for channel name.
    * @param callbacks Callback during subscription process.
-   * @param queue Creates queue for channel parallel to 'main'.
+   * @param queue Separate queue for channel's actions.
    */
   channel<
     ChannelParams extends string[] = string[],
--- a/base-server/index.d.ts
+++ b/base-server/index.d.ts
@@ -19,7 +19,7 @@ import { LogFn } from 'pino'
 
 import { Context, ChannelContext } from '../context/index.js'
 import { ServerClient } from '../server-client/index.js'
-import { Queue, queueChannel } from '../queue/index.js'
+import { Queue } from '../queue/index.js'
 
 export interface ServerMeta extends Meta {
   /**
@@ -690,15 +690,15 @@ export class BaseServer<
 
   /**
    * Types of queues.
-   * 
+   *
    * Queues for processing different actions in parallel.
    * By default every action is processed in 'main' queue.
    */
-  queueTypes: Map<string, queueChannel>
+  queueTypes: Map<string, string>
 
   /**
    * Queues by client ID and queue type.
-   * 
+   *
    * each queue processes actions from one client.
    */
   queues: Map<string, Queue>
@@ -985,10 +985,12 @@ export class BaseServer<
    *
    * @param name The action’s type or action’s type matching rule as RegExp..
    * @param callbacks Callbacks for actions with this type.
+   * @param queue Separate queue for actions withc this type.
    */
   type<TypeAction extends Action = AnyAction, Data extends object = {}>(
     name: TypeAction['type'] | RegExp,
-    callbacks: ActionCallbacks<TypeAction, Data, Headers>
+    callbacks: ActionCallbacks<TypeAction, Data, Headers>,
+    queue?: string
   ): void
 
   /**
@@ -1052,7 +1054,7 @@ export class BaseServer<
    *
    * @param pattern Pattern for channel name.
    * @param callbacks Callback during subscription process.
-   * @param queue Separate queue for channel actions.
+   * @param queue Separate queue for channel's actions.
    */
   channel<
     ChannelParams extends object = {},
@@ -1061,13 +1063,13 @@ export class BaseServer<
   >(
     pattern: string,
     callbacks: ChannelCallbacks<SubscribeAction, Data, ChannelParams, Headers>,
-    queue?: boolean
+    queue?: string
   ): void
 
   /**
    * @param pattern Regular expression for channel name.
    * @param callbacks Callback during subscription process.
-   * @param queue Creates separate queue for channel actions.
+   * @param queue Creates queue for channel parallel to 'main'.
    */
   channel<
     ChannelParams extends string[] = string[],
@@ -1076,7 +1078,7 @@ export class BaseServer<
   >(
     pattern: RegExp,
     callbacks: ChannelCallbacks<SubscribeAction, Data, ChannelParams, Headers>,
-    queue?: boolean
+    queue?: string
   ): void
 
   /**
--- a/base-server/index.d.ts
+++ b/base-server/index.d.ts
@@ -978,10 +978,12 @@ export class BaseServer<
   /**
    * @param actionCreator Action creator function.
    * @param callbacks Callbacks for action created by creator.
+	 * @param queue Processes action in queue.
    */
   type<Creator extends AbstractActionCreator, Data extends object = {}>(
     actionCreator: Creator,
-    callbacks: ActionCallbacks<ReturnType<Creator>, Data, Headers>
+    callbacks: ActionCallbacks<ReturnType<Creator>, Data, Headers>,
+		queue?: boolean
   ): void
 
   /**
--- a/base-server/index.d.ts
+++ b/base-server/index.d.ts
@@ -19,6 +19,7 @@ import { LogFn } from 'pino'
 
 import { Context, ChannelContext } from '../context/index.js'
 import { ServerClient } from '../server-client/index.js'
+import { Queue, queueChannel } from '../queue/index.js'
 
 export interface ServerMeta extends Meta {
   /**
@@ -688,6 +689,20 @@ export class BaseServer<
   connected: Map<string, ServerClient>
 
   /**
+   * Queue channels.
+   * Queues for processing different actions.
+   * Default queue has 'main' as a key, other queues can be
+   * created for different channels.
+   */
+  queueChannels: Map<string, queueChannel>
+
+  /**
+   * Queues by client ID.
+   * each queue processes actions from one client.
+   */
+  queues: Map<string, Queue>
+
+  /**
    * Connected client by client ID.
    *
    * Do not rely on this data, when you have multiple Logux servers.
@@ -978,12 +993,10 @@ export class BaseServer<
   /**
    * @param actionCreator Action creator function.
    * @param callbacks Callbacks for action created by creator.
-	 * @param queue Processes action in queue.
    */
   type<Creator extends AbstractActionCreator, Data extends object = {}>(
     actionCreator: Creator,
-    callbacks: ActionCallbacks<ReturnType<Creator>, Data, Headers>,
-		queue?: boolean
+    callbacks: ActionCallbacks<ReturnType<Creator>, Data, Headers>
   ): void
 
   /**
@@ -1038,6 +1051,7 @@ export class BaseServer<
    *
    * @param pattern Pattern for channel name.
    * @param callbacks Callback during subscription process.
+   * @param queue Separate queue for channel actions.
    */
   channel<
     ChannelParams extends object = {},
@@ -1045,12 +1059,14 @@ export class BaseServer<
     SubscribeAction extends LoguxSubscribeAction = LoguxSubscribeAction
   >(
     pattern: string,
-    callbacks: ChannelCallbacks<SubscribeAction, Data, ChannelParams, Headers>
+    callbacks: ChannelCallbacks<SubscribeAction, Data, ChannelParams, Headers>,
+    queue?: boolean
   ): void
 
   /**
    * @param pattern Regular expression for channel name.
    * @param callbacks Callback during subscription process.
+   * @param queue Creates separate queue for channel actions.
    */
   channel<
     ChannelParams extends string[] = string[],
@@ -1058,7 +1074,8 @@ export class BaseServer<
     SubscribeAction extends LoguxSubscribeAction = LoguxSubscribeAction
   >(
     pattern: RegExp,
-    callbacks: ChannelCallbacks<SubscribeAction, Data, ChannelParams, Headers>
+    callbacks: ChannelCallbacks<SubscribeAction, Data, ChannelParams, Headers>,
+    queue?: boolean
   ): void
 
   /**
--- a/base-server/index.js
+++ b/base-server/index.js
@@ -13,6 +13,7 @@ import { bindBackendProxy } from '../bind-backend-proxy/index.js'
 import { createHttpServer } from '../create-http-server/index.js'
 import { ServerClient } from '../server-client/index.js'
 import { Context } from '../context/index.js'
+import { Queue } from '../queue/index.js'
 
 const SKIP_PROCESS = Symbol('skipProcess')
 const RESEND_META = ['channels', 'users', 'clients', 'nodes']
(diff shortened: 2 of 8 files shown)


queue/index.d.ts
```diff
@@ -0,0 +1,34 @@
+import { Action, Meta } from '@logux/core'
```
ai, line 1: It will not be a public API, we can remove types

queue/index.js
```diff
@@ -0,0 +1,47 @@
...
+          await this.app.log.add(action, meta).then(() => this.remove())
+      )
+  }
+
+  waitForProcess() {
+    return new Promise(resolve => {
+      if (this.processing) {
+        setTimeout(this.waitForProcess, 50)
+      } else {
+        resolve(`Queue ${this.key} finished processing`)
```
ai, line 43: Do we have any reason to use string here?

base-server/index.js
```diff
@@ -316,10 +317,12 @@ export class BaseServer {
     this.unbind = []
 
     this.connected = new Map()
+    this.queues = new Map()
     this.nodeIds = new Map()
     this.clientIds = new Map()
     this.userIds = new Map()
     this.types = {}
+    this.queueTypes = []
```
ai, line 325: Why do we need this array?

queue/index.js
```diff
@@ -0,0 +1,47 @@
...
+  constructor(app, actionTypes, key) {
+    this.app = app
+    this.actionTypes = actionTypes
+    this.key = key
+    this.data = []
+    this.processing = false
+  }
+
+  add(action, meta) {
+    this.data.push(action)
```
ai, line 11: Let’s add them together as `this.data.push([action, meta])`

queue/index.js
```diff
@@ -0,0 +1,47 @@
...
+    this.data = this.data.slice(2)
+    if (this.data.length) {
+      this.processAction(this.data[0], this.data[1])
+    } else {
+      this.processing = false
+    }
+  }
+
+  async processAction(action, meta) {
+    let client = this.app.connected.get(this.key)
```
ai, line 29: What if action will be processed when client will be already disconnected

1. Client A connects
2. Client B connects
3. Client A adds action 1
4. Queue is busy with action 1 processing
5. Client B adds action 2
6. Client B disconnects
7. Queue finished action 1 processing
8. Queue go to action 2, but Client B is already disconnected

server-client/index.js
```diff
@@ -204,6 +204,12 @@ export class ServerClient {
       return true
     }
 
+    if (this.app.queueTypes.includes(type) && !queueCall) {
+      let queue = this.app.queues.get(this.key)
```
ai, line 208: Note that we should have separated queue for each queue and client https://github.com/logux/logux/issues/54

> For every client and queue, we are creating action queue. 

Here you put all client to the same queue

base-server/index.js
muddv: `queueTypes` holds types of actions to be processed in a queue.  This array is checked in `ClientServer.filter` to see if action should be processed normally or in queue and it is populated if the developer adds true when declaring `type` in a server module.

queue/index.js
muddv: Yes this would be better!

queue/index.js
muddv: This should not cause any problems. As far as I understand, the `connected` Map in `BaseServer` does not get destroyed until the server itself is turned off. So this is what should happen in this scenario:
After step 8, the action is probably already marked as `synced` (or, otherwise not in queue at all) as everything up to `ServerClient.filter` happens in parallel, not in queue, the client will not be re-sending this action again if reconnected.
Similarly to the `connected` map in `BaseServer`, the `queues` map is not destroyed until server shuts down, so the queued action will be added to the log even if the client is already disconnected.

server-client/index.js
muddv: When a new client is connected, a new queue is declared for it and put in a map in `BaseServer.addClient`. So, the queue in this function should be a particular queue, belonging to this client.

queue/index.d.ts
muddv: Ok, hopefully they were at least a bit helpful, when looking at this pr :smile:

queue/index.js
muddv: No good reason, I will remove it

queue/index.js
muddv: Maybe the `client` variable here should be renamed to `serverClient` for clarity?

ai: > As far as I understand, the connected Map in BaseServer does not get destroyed until the server itself is turned off

The problem is that `ServerClient` remove itself from `server.connected` on client’s disconnect

https://github.com/logux/server/blob/main/server-client/index.js#L99

base-server/index.js
ai: Why not `Map` with `type → queue name`?

queue/index.js
muddv: I see now. Maybe I can add a check to a `destroy` function, so that it would wait for queue to complete, or maybe it would be even better to rewrite `filter` function in `queue.processAction` (from this line: https://github.com/logux/server/blob/main/server-client/index.js#L207).
Also, I should add a way to delete a queue if a respective client is disconnected and all actions are processed.

ai: >  I can add a check to a destroy function, so that it would wait for queue to complete

`server.connected` is used to detect current active clients. We can’t change behavior.

We should be ready to process any action without a `ServerClient` instance. What if server will add its own action by `server.log.add(action, meta)`?

base-server/index.js
muddv: `Map` would make a lot sense here. But are there any clear advantages over an array?

ai: > But are there any clear advantages over an array?

I do not see an advantage of array. Every action has some queue (at least `main`).

queue/index.js
muddv: Ok, I will add ability to process action without `ServerClient`, thank you for the pointers

base-server/index.js
```diff
@@ -630,6 +643,10 @@ export class BaseServer {
     let key = this.lastClient.toString()
     let client = new ServerClient(this, connection, key)
     this.connected.set(key, client)
+    if (this.queueTypes.length) {
```
ai, line 646: Even if all actions uses `main` queue, actions need to be processed with a user’s queue. Why do we have here a `if`?

muddv: I thought it might be beneficial to add this functionality in as non invasive way as possible, so if the developer decides that all potential actions are safe, queues can be neglected entirely as there are no actions that should be processed in queue and everything is presumably safe to process in parallel
That is also why I used array instead of `Map` for `queueTypes`. I can rework my solution to have a `main` queue.
Does it mean that actions in the `main` queue, should be processed in parallel, or should the queue be default way to process actions with ability to opt out for certain types?

ai: Yes, unfortunately, we can’t add a new feature, we need to refactor the old one.

Old logic (all tasks are processing in parallel) leads to painful issues (like `logux/subscribe` to post `A` was executed before `posts/create` A was not finished).

muddv: I understand, I will rework my solution to support this

muddv:
Now all of the client's actions are now processed through queues.
- I added queues to `server-client`'s test, because test because test clients there do not are not created through normal `BaseServer.addClient` method, causing them to fail. 
- I left some type definitions for queues, because test files are `.ts` and they require types, especially in `server-client` tests, also some parts of queues interact with typed and public parts of `BaseServer` and `ServerClient`, but maybe some defs can be left out or converted to `any` to not confuse this with a public API.

queue/index.js
muddv: Done

queue/index.js
muddv: Fixed

muddv:
Also `notifies about actions and subscriptions` test in `bind-backend-proxy` (https://github.com/logux/server/blob/main/bind-backend-proxy/index.test.ts#L438) now fails, this has to do with order in which actions are now processed, but i decided to not change it right now in case this could be actually detrimental in some way. Anyway, at the end test log contains all the sent messages, but in different order and with different `added` index.

ai:
Can you try to fix `notifies about actions and subscriptions` test? It really could be a result of order changes and bad code in that test (but let’s be sure).

base-server/index.js
```diff
@@ -496,6 +505,10 @@ export class BaseServer {
       channel.regexp = pattern
     }
     this.channels.push(channel)
+
+    if (queue) {
+      this.queueChannels.set(pattern, queueChannel)
```
ai, line 510: Seems like you are copying the same `queueChannel` instance to every query.

Do we need to create a new instance here?

queue/index.js
```diff
@@ -0,0 +1,122 @@
...
+
+export class Queue {
+  constructor(app, queues, key) {
+    this.app = app
+    this.queues = queues
+    this.key = key
+    this.clientId = undefined
+  }
+
+  setClientId(clientId) {
```
ai, line 14: Why do need to change client ID in runtime? Will it be more readable to have immutable client ID and create new `Queue` instance for every client?

base-server/index.js
```diff
@@ -316,10 +317,12 @@ export class BaseServer {
     this.unbind = []
 
     this.connected = new Map()
+    this.queues = new Map()
     this.nodeIds = new Map()
     this.clientIds = new Map()
     this.userIds = new Map()
     this.types = {}
+    this.queueChannels = new Map().set('main', queueChannel)
```
ai, line 325: Will it be more readable to have flat `Map<queue + clientId, Queue>` here instead of more complex nested system?

base-server/index.js
muddv: Yes, that is an oversight on my part. I replaced object with constructor.

queue/index.js
muddv: Problem is `ServerClient` also starts with client id as `undefined` (https://github.com/logux/server/blob/main/server-client/index.js#L21). Right now queue gets client id from its respective client in `ServerClient.auth` method (https://github.com/muddv/logux-server/blob/queues/server-client/index.js#L109). Is this problematic?

base-server/index.js
muddv: We can simplify it. Right now `BaseServer.queues` (https://github.com/muddv/logux-server/blob/queues/base-server/index.js#L320) is such a flat map, it  contains all current client's queues and  `queueChannels` is just needed to hold information about which actions can be processed in parallel.

ai: Actions who every client should be processed in parallel, so why do we need `queueChannels` with per-queue map?

muddv: `queueChannels` are meant to be used by one client to process different actions in parallel, if the developer allows it during the creation of channel (https://github.com/muddv/logux-server/blob/queues/base-server/index.js#L509).
So for example, if server has 2 channels, like `users/` and `posts/`, Queue could be processing 2 actions from them, if needed.

queue/index.js
```diff
@@ -0,0 +1,124 @@
...
+  add(action, meta) {
+    let queueKey
+    if (this.queues.size === 1) {
+      queueKey = 'main'
+    } else {
+      let labels = Array.from(this.queues.keys())
+      let type = action.type
+      let actionChannel = type.slice(0, type.indexOf('/'))
+      if (labels.includes(actionChannel)) {
+        queueKey = actionChannel
```
ai, line 30: We should not use channels as `queueKey`.

For instance, if we have `posts` and `comments` channels, their actions should not be processed in parallel. Otherwise, we can try to create a comment before comment’s post was created.

queue/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+  setClientId(clientId) {
+    this.clientId = clientId
+    return this.clientId
+  }
+
+  add(action, meta) {
+    let queueKey
+    if (this.queues.size === 1) {
+      queueKey = 'main'
```
ai, line 24: `queueKey` should be `main` only if developer didn’t pass any queue name in `Server#type()`. It should not be dependent from number of defined queues.

base-server/index.js
```diff
@@ -666,6 +680,36 @@ export class BaseServer {
     this.debugActionError(meta, `Wrong channel name ${action.channel}`)
   }
 
+  onAction(action, meta, clientId) {
+    let queueType
+    if (!this.queueTypes[0]) {
+      queueType = 'main'
+    }
+    else {
+      let type = action.type.slice(0, action.type.indexOf('/'))
```
ai, line 689: We should not use `/` magic:

1. It is not part of Logux Server semantic. For server, every `action.type` is just a string.
2. We are making DX more complicated. It is better to explicitly define `queue` for each type, rather of explaining complex magic logic.

server-client/index.js
```diff
@@ -201,34 +201,18 @@ export class ServerClient {
...
-        return true
-      }
-    } catch (e) {
-      this.app.undo(action, meta, 'error')
-      this.app.emitter.emit('error', e, action, meta)
-      this.app.finally(processor, ctx, action, meta)
-      return false
-    }
+    this.app.onAction(action, meta, this.clientId)
+    return false
```
ai, line 215: Not really like that, we return always `false` and re-add actions later. It looks like too hacky.

Can `this.app.onAction()` return `Promise<boolean>`?

queue/index.js
```diff
@@ -0,0 +1,85 @@
...
+      }
+    } catch (e) {
+      this.app.undo(action, meta, 'error')
+      this.app.emitter.emit('error', e, action, meta)
+      this.app.finally(processor, ctx, action, meta)
+      this.remove()
+    }
+  }
+
+  remove() {
```
ai, line 52: What do you think about `next()` name?

base-server/index.js
muddv: What would be a good way to place `logux/` actions into a correct queue?
If developer creates a queue for a particular type, but not entire channel, we may not have enough information.  One way I see is to store action ids to reliably place `logux/subscribe` to correct queue, but to me it seems it would require even more complex logic.

ai: User will define queue for channels: `server.channel(…, { queue: 'users' })`

If action processing, we will check `action.type === 'logux/subscribe || …` explicitly and use `action.channel` to get `queue` from `server.channel` definitions.

server-client/index.js
muddv: Should we return `this.app.onAction()` from filter?
We can change `onAction` to return a Promise, but we will have to resolve it to `false` unless we also change the `core`, because `syncMessage` would add duplicate action to log (https://github.com/logux/core/blob/main/sync/index.js#L89). Would it be ok to resolve it to `false`?

ai: > because syncMessage would add duplicate action to log

I mean remove `this.app.log.add()` from queue’s code and return `Promise<true|false>` instead from queue.

base-server/index.js
muddv: Should we allow user to place a particular `type` to a separate queue? If such ability is not needed, then this is simple

ai: > Should we allow user to place a particular type to a separate queue?

Yes, each `type` must explicitly has queue. No auto-generation for queue because queues should be explicit.

server-client/index.js
muddv: I `onAction` to return promise, but currently it may be problematic for `sendSynced in `core`, maybe some changes can be made in `core/sync` to synchronize actions faster witch queues.

queue/index.js
muddv: Good idea!

muddv:
Changing `BaseServer.onAction` to return promise from queue inexplicably caused `notifies about actions and subscriptions test` in `bind-backend-proxy` to break once again. It does not seem to be detrimental, but maybe it is worth to look into.

ai:
Sorry, I am in the travel. I hope to review changes on this week.

base-server/index.js
```diff
@@ -375,6 +379,10 @@ export class BaseServer {
           }
         })
     )
+
+    this.unbind.push(() => {
+      this.queues.forEach(async i => await i.waitForProcess())
```
ai, line 384: You need to return `Promise.all()` here to wait until the end

base-server/index.d.ts
```diff
@@ -687,6 +688,21 @@ export class BaseServer<
    */
   connected: Map<string, ServerClient>
 
+  /**
+   * Types of queues.
```
ai, line 692: Let’s not make them public

ai: I do not want to think about backward compatibility of them yet

base-server/index.js
```diff
@@ -496,6 +509,16 @@ export class BaseServer {
       channel.regexp = pattern
     }
     this.channels.push(channel)
+
+    if (queue) {
+      channel.regexp
+        ? this.queueTypes.set(channel.regexp, queue)
+        : this.queueTypes.set(channel.pattern.regex)
```
ai, line 516:
```suggestion
        : this.queueTypes.set(channel.pattern.regex, queue)
```

base-server/index.js
```diff
@@ -496,6 +509,16 @@ export class BaseServer {
       channel.regexp = pattern
     }
     this.channels.push(channel)
+
+    if (queue) {
+      channel.regexp
+        ? this.queueTypes.set(channel.regexp, queue)
```
ai, line 515: Channel could have the same name as some `action.type`. We need a prefix here or a different `Map`.

server-client/index.js
ai: What problem of `sendSynced` do you mean? It should be sent to client when the server will add it to the own log (and server need to call `access` first). So we need to wait the queue first.

muddv: This is not a huge concern, but I meant that `sendSynced` for action can take a lot of time if client sends several actions at once.

Mordorreal:
@ai Are current solution will be merged? Does it need additional work? I have time now and can try to finish this, many years' effort.

muddv:
@Mordorreal my solution definitely still need some tweaks, but I think I can restart working on it, depending on what @ai thinks here. From my pov, if you can provide a clean solution right now, you can try to go ahead, again depending on what maintainer decides

ai:
This PR can be finished, but we need to replace how we work with actions in `BaseNode` (see the discussion above)

PR #141, opened 2023-05-24
Multiple subscriptions with filters per node

euaaaio:
Looks like we had a bug. When client is subscribed to the same channel multiple times, but with different filters, he no longer receives resent finished actions.

With this code, the client will only receive `posts/deleted` actions:
```js
let postsA = useFilter(PostsStore, { category: 'A' })
let postsB = useFilter(PostsStore, { category: 'B' })
```

With this PR we can subscribe on the same channel multiple times with different filters.
I used hashmap to collect filters so that it was easier to remove them on unsubscribe.

### DIFF
--- a/base-server/index.js
+++ b/base-server/index.js
@@ -94,6 +94,10 @@ function normalizeChannelCallbacks(pattern, callbacks) {
   }
 }
 
+function subscriberFilterId(action) {
+  return JSON.stringify(action.filter || {})
+}
+
 export class BaseServer {
   constructor(opts = {}) {
     this.options = opts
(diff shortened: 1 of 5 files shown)


base-server/index.js
```diff
@@ -593,23 +597,23 @@ export class BaseServer {
...
-                if (typeof filter === 'function') {
-                  if (!ctx) ctx = this.createContext(action, meta)
-                  filter = await filter(ctx, action, meta)
-                }
+                let ctx = this.createContext(action, meta)
                 let client = this.clientIds.get(clientId)
-                if (filter && client) {
-                  ignoreClients.add(clientId)
-                  client.node.onAdd(action, meta)
+                for (let [, filter] of Object.entries(subscriber.filters)) {
```
ai, line 609:
```suggestion
                for (let filter of Object.values(subscriber.filters)) {
```

ai:
I will release it after Logux Client migration to Nano Stores 0.9.

ai:
Released at 0.12.3.

PR #145, opened 2023-08-05
fix(base-server): broken processing with async filter

euaaaio:


ai:
Thanks. Released at 0.12.5.

PR #146, opened 2023-08-06
Add action processing queues

VladBrok:
Closes https://github.com/logux/logux/issues/54
Depends on https://github.com/logux/core/pull/52 (will not work properly until https://github.com/logux/core/pull/52 is merged)

This PR adds action processing queues.

A developer can specify the name of the queue for different actions and channels using Server#type and Server#channel.
If the name is not specified, the default 'main' will be used.
If a developer wants to run actions in parallel, they can add them to different queues.

Queues are created for each client, for example, if a client with id '10' creates the 'main' queue and a client with id '11' creates the 'main' queue, these would be different queues even though they have the same name, because the client ids are different.

Queues for channels and actions are always different, even if they have the same name.

if an error occurs during action processing, the queue will undo all it's other actions.


package.json
```diff
@@ -37,6 +37,7 @@
     "JSONStream": "^1.3.5",
     "cookie": "^0.5.0",
     "dotenv": "^16.3.1",
+    "fastq": "^1.15.0",
```
ai, line 40: You can test you latest version of `@logux/core` by using `"@logux/core": " VladBrokcore/#feat/54"` in `package.json`

base-server/index.d.ts
```diff
@@ -716,6 +737,8 @@ export class BaseServer<
    */
   nodeIds: Map<string, ServerClient>
 
+  onSync: SyncCallback
```
ai, line 740: `meta.*` should be serializable. You can put functions to `meta`.

base-server/index.d.ts
```diff
@@ -1105,6 +1132,16 @@ export class BaseServer<
...
 
+  /**
+   * Sets the name of the queue that will be used to process actions of this type.
+   * Can be used for actions which type was not defined
+   * by any {@link Server#type}.
+   *
+   * @param actionType the action type
+   * @param queue the queue name
+   */
+  setQueue(actionType: string, queue: string): void
```
ai, line 1143: Why we want to use this and not `server.type()`?

server-client/index.js
```diff
@@ -35,6 +35,7 @@ export class ServerClient {
       auth: this.auth.bind(this),
       inFilter: this.filter.bind(this),
       inMap: this.inMap.bind(this),
+      onSync: app.onSync?.bind(app),
```
ai, line 38:
```suggestion
      onSync: app.onSync.bind(app),
```

Seems like it always exists

base-server/index.js
```diff
@@ -373,6 +423,16 @@ export class BaseServer {
           }
         })
     )
+    this.unbind.push(() => {
+      let queues = [...this.queues.values()]
+      let promises = queues.map(
```
ai, line 428: Let’s combine these three steps. In this specific case it will simplify readability (because we need `Promise.allSettled` to see first to understand what we are doing.

```js
return Promise.allSettled(
  [...this.queues.values()].map(queue => {
    return new Promise(resolve => {
      queue.drain = resolve
    })
  })
)
```

base-server/index.js
```diff
@@ -420,6 +480,10 @@ export class BaseServer {
       channel.regexp = pattern
     }
     this.channels.push(channel)
+
+    let queue = options?.queue || 'main'
+    let channelPattern = channel.regexp ? channel.regexp : channel.pattern.regex
+    this.channelPatternToQueueName.set(channelPattern, queue)
```
ai, line 486: Can we use `this.channels` instead? Anyway is an array of objects, where we can add new keys.

base-server/index.js
```diff
@@ -333,6 +378,11 @@ export class BaseServer {
     this.timeouts = {}
     this.lastTimeout = 0
 
+    this.channelPatternToQueueName = new Map()
+    this.actionTypeToQueueName = new Map()
```
ai, line 382: Let’s use `this.typeToQueue` to keep origin short names.

base-server/index.js
```diff
@@ -617,6 +681,51 @@ export class BaseServer {
...
+      queueName = this.actionTypeToQueueName.get(action.type)
+    }
+
+    queueName = queueName || 'main'
+
+    let queueKey = `${clientId}/${queueName}`
+    let queue = this.queues.get(queueKey)
+
+    if (!queue) {
+      let concurrency = 1
```
ai, line 715: I am not sure that we need this variable, just call `queue = fastq(queueWorker, 1)`

Some developer will not understand what `1` means, but:
1. We don’t work with concurrency here
2. They have types for `fastq`

base-server/index.js
```diff
@@ -64,6 +65,48 @@ function normalizeTypeCallbacks(name, callbacks) {
   }
 }
 
+function queueWorker(arg, cb) {
```
ai, line 68: Let’s call `cb` like `end` or `next` to explain what it there

base-server/index.js
```diff
@@ -64,6 +65,48 @@ function normalizeTypeCallbacks(name, callbacks) {
...
+  let { action, meta, queueKey, server } = arg
+  let queue = server.queues.get(queueKey)
+
+  let end = (...args) => {
+    server.actionsInQueue.delete(meta.id)
+    if (!queue.length()) {
```
ai, line 74:
```suggestion
    if (queue.length() === 0) {
```

base-server/index.js
```diff
@@ -64,6 +65,48 @@ function normalizeTypeCallbacks(name, callbacks) {
...
+  }
+
+  let undoRemainingTasks = () => {
+    for (let task of queue.getQueue()) {
+      server.undo(task.action, task.meta, 'error')
+    }
+    queue.killAndDrain()
+  }
+
+  let unbindError = server.on('error', (e, errorAction) => {
```
ai, line 87: It is not very effective to bind new listener on every action (you could have 500 of them in a second).

base-server/index.d.ts
VladBrok: Sorry, I don't understand. The `onSync` is a method in `BaseServer` and it is not saved to the `meta`.

base-server/index.d.ts
ai: Nope, let’s use `server.type()`. `server.otherType()` is for proxy, they will need a different approach for queue anyway.

base-server/index.d.ts
ai: Sorry, my fall. I was confused by diff (thought that it is an addition to `ServerMeta`.

base-server/index.js
```diff
@@ -409,7 +471,7 @@ export class BaseServer {
     return [undoAction, undoMeta]
   }
 
-  channel(pattern, callbacks) {
+  channel(pattern, callbacks, options) {
```
ai, line 474:
```suggestion
  channel(pattern, callbacks, options = {}) {
```

It will allow to replace `options?.queue || 'main'` to `options.queue || 'main'`

base-server/index.js
```diff
@@ -355,6 +369,45 @@ export class BaseServer {
     this.listenNotes = {}
     bindBackendProxy(this)
 
+    let end = (meta, queue, queueKey, ...args) => {
```
ai, line 372:
```suggestion
    function end (meta, queue, queueKey, ...args) {
```

A small style fix to keep code style consistent

base-server/index.js
```diff
@@ -355,6 +369,45 @@ export class BaseServer {
...
+      this.actionToQueue.delete(meta.id)
+      if (queue?.length() === 0) {
```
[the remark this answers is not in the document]
VladBrok: Good question, it should never be missed at this point, but it was. 

The issue was that inside `on('processed')`, I determine whether an action belongs to the queue by using `actionToQueue` map. 
I save an action to this map by `meta.id`, but inside `on('processed')` I lookup the map by `meta.id` and `action.id` (because I may catch `logux/undo`, which has different `meta.id` but might have `action.id` of some action in the queue, in which case I need to undo all other actions in that queue). 

However, after processing, I always deleted an action from the `actionToQueue` by `meta.id`, so if it was `logux/undo` action, the following happened:

1. `on('processed')` got `logux/undo` with `action.id = '1'` and `meta.id = '2'`
2. `actionToQueue` has an action with `meta.id = '1'` equal to `action.id = '1'` so we process `logux/undo`
3. we undo all remaining actions in this queue and clear the queue
4. we try to delete this action (`logux/undo`) from the `actionToQueue` map by `meta.id = '2'` and delete nothing
5. the queue is deleted because it is empty (all actions are undone and removed from it)
6. `on('processed')` got some other action with `meta.id = '1'`
7. `actionToQueue` has an action with `meta.id = '1'` so we process it
8. we try to use the queue which was deleted in step `5`

Instead, in step `4` we should have deleted an action from `actionToQueue` using `action.id` so that in step `7` an action would be ignored because it was undone.

base-server/index.d.ts
VladBrok: maybe we should disable queues for proxy? Right now, if I remove `setQueue`. test 'reacts on wrong backend answer' will fail because all actions will by default go to `main` queue, but the test assumes that actions are executed in parallel

ai: Can we change this test? We will release it in major update.

base-server/index.js
```diff
@@ -355,6 +368,46 @@ export class BaseServer {
...
+      if (!queueKey) {
+        return
+      }
+      let queue = this.queues.get(queueKey)
+      undoRemainingTasks(queue)
+      end(meta.id, queue, queueKey, e)
+    })
+    this.on('processed', (action, meta) => {
+      let queueKey =
+        this.actionToQueue.get(meta?.id) || this.actionToQueue.get(action?.id)
```
ai, line 399: Why do we need `this.actionToQueue.get(action?.id)`?

ai: Do we need it for `logux/undo` actions? Let’s put `this.actionToQueue.get(action?.id)` in `if (action.type === 'logux/undo')` branch to make it clear and avoid bugs if some other action will have `action.id`.

VladBrok: Actions are saved to the `actionToQueue` by `meta.id`. `on('processed')` will receive `logux/processed` which [per docs](https://logux.org/guide/concepts/action/#system-actions) has `action.id` equal to the `meta.id` of the processed action.

Actually we don't need `this.actionToQueue.get(meta?.id)` here.

base-server/index.js
```diff
@@ -355,6 +372,45 @@ export class BaseServer {
...
+      let queueKey = this.actionToQueue.get(meta?.id)
+      if (!queueKey) {
+        return
+      }
+      let queue = this.queues.get(queueKey)
+      undoRemainingTasks(queue)
+      end(meta.id, queue, queueKey, e)
+    })
+    this.on('processed', (action, meta) => {
+      let queueKey = this.actionToQueue.get(action?.id)
```
ai, line 402: You can’t check `action.id` until you check `action.type`. Diferent actions could have a very dfferent data in `action` with possible collisions.

So add `logux/processed` or something here. But I think we should check `meta.id` to mark tasks as processed on their own `processed` event, not on `processed` event of `logux/processed` action for that action.

PR #148, opened 2023-08-15
fix(add-sync-map-filter): action filter types and built-in filter

euaaaio:
- Fixed types of resent actions
- Don’t resend `changed` actions that didn’t pass filter

### DIFF
--- a/add-sync-map/index.js
+++ b/add-sync-map/index.js
@@ -44,6 +44,14 @@ function buildFilter(filter) {
         if (action.fields[key] !== filter[key]) return false
       }
     }
+    if (action.type.endsWith('/changed')) {
+      for (let key in filter) {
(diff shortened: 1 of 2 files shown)


add-sync-map/index.js
```diff
@@ -44,6 +44,14 @@ function buildFilter(filter) {
         if (action.fields[key] !== filter[key]) return false
       }
     }
+    if (action.type.endsWith('/changed')) {
+      for (let key in filter) {
+        if (
+          action.fields[key] &&
```
ai, line 50:
```suggestion
          key in action.fields &&
```

Otherwise if will not work with `""` and `false`.

ai:
Thanks. Released at 0.12.6.

